PR #11, opened 2017-03-31
Include <cerrno>

ppwwyyxx:
When compiling on Linux host to armv7a target following to the tutorial, got error of "errno undeclared". Include this header fixed this problem.


llhe:
+1

AnthonyARM:
Thanks for fixing this issue !

PR #114, opened 2017-05-18
Add a build flag to control whether or not the example programs get built

andrew-wja:
At present, compilation of some example programs fails, causing the build of the library to fail. However, the library itself builds fine, and the problem seems to be with the example programs. This changeset adds the flag examples=(yes|no) to scons, so that the library can be built without also building the example programs. In addition, some files and directories which are created automatically by scons have been added to the .gitignore so that git does not try to add built objects to commits.

### DIFF
--- a/sconscript
+++ b/sconscript
@@ -351,30 +351,32 @@ alias = env.Alias("arm_compute",objects)
 Default(alias)
 
 # Build examples
-test_helpers = env.Object("test_helpers/Utils.cpp")
 
-if env['opencl'] and env['neon']:
-    for file in Glob("examples/neoncl_*.cpp"):
-        example =  os.path.basename( os.path.splitext(str(file))[0])
-        prog = env.Program(example, ['examples/%s.cpp' % example, test_helpers], LIBS=example_libs+['OpenCL'])
-        alias = env.Alias(example, prog)
-        Depends(prog, objects)
-        Default( alias )
+if env['examples']:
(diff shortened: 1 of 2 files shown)


AnthonyARM:
If the examples fail to build could you please open an issue with more details ?
Thanks

andrew-wja:
Sure, I'll create an issue with the details.

.gitignore
```diff
@@ -0,0 +1,2 @@
+build/
+.sconsign.dblite
```
AnthonyBarbier: Please remove that file from the pull request

sconscript
```diff
@@ -351,30 +351,32 @@ alias = env.Alias("arm_compute",objects)
...
-
-if env['neon']:
-    for file in Glob("examples/neon_*.cpp"):
-        example =  os.path.basename( os.path.splitext(str(file))[0])
-        prog = env.Program(example, ['examples/%s.cpp' % example, test_helpers], LIBS=example_libs)
-        alias = env.Alias(example, prog)
-        Depends(prog, objects)
-        Default( alias )
+if env['examples']:
+  test_helpers = env.Object("test_helpers/Utils.cpp")
```
AnthonyBarbier: Use same level of indentation as the rest of the file (4 spaces per level)

andrew-wja:
Hi, I've fixed the indentation and squashed the spurious commits, sorry about that!

PR #199, opened 2017-08-10
Use benchmark.h instead of deprecated benchmark_api.h

JammyZhou:


GeorgeARM:
As we removed dependency from google benchmark I am closing this request as resolved.
Reopen if needed.

PR #325, opened 2018-01-02
ARM_COMPUTE_NO_EXCEPTIONS macro guard

lukeiwanski:
Adds ARM_COMPUTE_NO_EXCEPTIONS macro guard for targets that do not accept exceptions


lukeiwanski:
@GeorgeARM is this better?

utils/Utils.h
```diff
@@ -24,6 +24,16 @@
 #ifndef __UTILS_UTILS_H__
 #define __UTILS_UTILS_H__
 
+#ifdef ARM_COMPUTE_NO_EXCEPTIONS
```
AnthonyBarbier: This file is not used by the library: it's only needed for tests / examples, so you shouldn't need to update it ?

include/libnpy/npy.hpp
```diff
@@ -80,11 +80,12 @@ inline void read_magic(std::istream& istream, unsigned char& v_major, unsigned c
   istream.read(buf, magic_string_length+2);
 
   if(!istream) {
-    throw std::runtime_error("io error: failed reading file");
+    ARM_COMPUTE_THROW("io error: failed reading file");
```
AnthonyBarbier: This is a third party library, we need to upstream the changes in a cleaner way.

src/core/Error.cpp
```diff
@@ -56,6 +56,7 @@ void arm_compute::error(const char *function, const char *file, const int line,
     va_end(args);
     throw std::runtime_error(err.error_description());
 }
+
```
AnthonyBarbier: Unneeded change

utils/Utils.cpp
```diff
@@ -70,7 +70,6 @@ int run_example(int argc, const char **argv, example &func)
 {
     std::cout << "\n"
               << argv[0] << "\n\n";
-
     try
```
AnthonyBarbier: Unneeded change

utils/Utils.h
lukeiwanski: Ah okay, so the really only changes are needed to npy.hpp?
It looked like this header is utility used by the framework.. obviously I was mistaken.
I will revert changes to this file.

include/libnpy/npy.hpp
lukeiwanski: Would that be acceptable to define `ARM_COMPUTE_THROW` in this file? Or do you have something else in mind?

lukeiwanski:
@GeorgeARM  seems like this https://github.com/ARM-software/ComputeLibrary/pull/325#commitcomment-26593339 will not be needed after all

include/libnpy/npy.hpp
AnthonyBarbier: Are you sure this is needed ? 
I just had a quick look and it looks like this is only included in `utils/Utils.h` and `tests/AssetsLibrary.h` ?
So I think if you build with `examples=0 benchmark_tests=0 validation_tests=0` then you should be fine.

utils/Utils.h
AnthonyBarbier: We have a different `arm_compute/core/Utils.h` but I don't think `utils/Utils.h` is used outside of the examples and tests

include/libnpy/npy.hpp
lukeiwanski: actually it's not needed!
My problem started with wrong assumptions that I need to include `utils/Utils.h`.
It turns out that the only change needed is that to `arm_compute/core/CL/OpenCL.h`

utils/Utils.h
lukeiwanski: yes, that essentially was cause of my confusion. I can confirm that `utils/Utils.h` is not needed!
I will revert not needed changes.

PR #343, opened 2018-01-29
Bugfix: must zero all values in all slices

wr0112358:
Only first 2d slice was zeroed out leading to wrong convolution results. Window method might not be the most efficient?


wr0112358:
The changed std::fill_n parameters solve the problem. I updated the commit accordingly.

PR #377, opened 2018-03-02
Fix links on README

gwtnb:
Fix links on README.


AnthonyARM:
Thanks for fixing the links! (We've done it too as part of the 18.03 update, so I won't merge that pull request)

PR #395, opened 2018-03-23
Bugfix: uninitialised buffer access in NEGEMMConvolutionLayer can lea…

wr0112358:
…d to invalid results

 I implemented a vgg16-fcn8 model that worked fine with 18.01, but predicts nan's since 18.02.
Finally time for debugging it: 1x1 NEON Convolution Layer, utilising NEGEMMConvolutionLayer inside access uninitialised memory leading to invalid results.

### DIFF
--- a/src/runtime/NEON/functions/NEGEMMConvolutionLayer.cpp
+++ b/src/runtime/NEON/functions/NEGEMMConvolutionLayer.cpp
@@ -614,6 +614,8 @@ void NEGEMMConvolutionLayer::run()
     // Run input reshaping
     NEScheduler::get().schedule(&_input_im2col_kernel, Window::DimY);
 
+    std::fill_n(_gemm_output.buffer(), _gemm_output.info()->total_size(), 0);
+
     // Runs matrix multiply on reshaped matrices
     if(_mm_optimised_kernel != nullptr)
     {


src/runtime/NEON/functions/NEGEMMConvolutionLayer.cpp
```diff
@@ -614,6 +614,8 @@ void NEGEMMConvolutionLayer::run()
     // Run input reshaping
     NEScheduler::get().schedule(&_input_im2col_kernel, Window::DimY);
 
+    std::fill_n(_gemm_output.buffer(), _gemm_output.info()->total_size(), 0);
```
gmiodice, line 617: Could you tell us which platform are you using?

wr0112358: arm64 qemu(qemu-system-aarch64 -cpu cortex-a57) for this test. The problem also occurred on a hikey960.

AnthonyARM:
We don't seem to be able to reproduce the issue, could you please share a reproducer case ?
We've got a demo running on Hikey 960 that runs multiple times VGG16 and it seems to work as expected.

wr0112358:
Thank you for the quick response.
It seems the nan problem on hikey960 is different to the qemu one. My first convolution layer(NEConvolutionLayer) has padding=100. The default used winograd convolution does not support this padding and fails silently. Replacing the first convolution layer with NEGEMMConvolutionLayer fixes the problem on hikey960 for me.

On qemu the nan problem persists and above patch fixes it. But qemu was just a stop-gap solution for me anyway.

AnthonyARM:
If you compile with `asserts=1` you should hopefully have more information.
It should only silently fail in release mode.

PR #399, opened 2018-03-29
Fix race condition on info.thread_id in OpenMP scheduler

simpel01:
Additionally refactor improper use of OpenMP's worksharing
construct (for).


AnthonyARM:
Thanks !

Issue #402, opened 2018-03-31
NEDeconvolutionLayer produces wrong results for higher ofm size


wr0112358:
**Output of 'strings libarm_compute.so | grep arm_compute_version':**
arm_compute_version=v18.03 Build options: {'arch': 'arm64-v8a', 'opencl': '1', 'neon': '1', 'asserts': '0', 'debug': '0', 'os': 'android', 'Werror': '1'} Git hash=c3f34a43ffb5d52ee4a4e9f7b1bf4c6c002aeebb
arm_compute_version=v18.03 Build options: {'arch': 'arm64-v8a', 'opencl': '1', 'neon': '1', 'asserts': '1', 'debug': '1', 'os': 'android', 'Werror': '1'} Git hash=c3f34a43ffb5d52ee4a4e9f7b1bf4c6c002aeebb

**Platform:**
hikey960
**Operating System:**
android 8.0

**Problem description:**
NEDeconvolutionLayer produces wrong results for higher ofm size. Parameters used:
Data Type: F32
Input shape: 18x16x1024
Weights shape: 4x4x1024x512
Output shape: 38x34x512
Bias: nullptr
arm_compute::PadStrideInfo(2, 2, 3, 3, arm_compute::DimensionRoundingType::FLOOR)
inner_border_right = inner_border_top = 0
kernel size: 4

Weights are generated with the following function:
https://github.com/MarvinTeichmann/tensorflow-fcn/blob/master/fcn8_vgg.py#L284
get_deconv_filter([ 4, 4, 512, 1024])
tf.nn.conv2d_transpose(bottom, weights, output_shape, strides=[1, 2, 2, 1], padding='VALID')

**Expected Result:**
- produce same result as tensorflow version. This is working for smaller ofm sizes.

If necessary I can provide .npy weights and code.

wr0112358:
Indeed, that is the case. This also explains why the ofm=ifm case works.
Just to be clear:
Until now i export my tensorflow weights with a .transpose(3,2,0,1)
If I understand this correctly, I have to export the weights for deconv-layer with: .transpose(2,3,0,1)?

PR #404, opened 2018-04-04
Nedeconv bugfix

wr0112358:
As described in issue 402, NEDeconvolutionLayer fails for a parametrisation with ifm != ofm.
"fix ofm, ifm mixup in deconvolution_output_shape" fixes a bug that mixes up ifm and ofm in output size calculation, this happened most likely due to the swapped weights dimensions for deconvolution layer.

The Rest of the changes adapts error handling to the ifm != ofm use case.


mpekatsoula:
Hi @wr0112358 ,
The issue here is that the API for Deconvolution accepts weights in [W, H, OFM, IFM] format while Convolution accepts weights in  [W, H, IFM, OFM] format thus, we cannot pass them directly from DeConvolution to Convolution layer. We believe it makes sense to change the Deconvolution API to accept weights in [W, H, IFM, OFM] format. I think no changes are needed in the code, as deconvolution_output_shape is calculated as if the weights were in [W, H, IFM, OFM] format.

I'll close this pull request but feel free to update https://github.com/ARM-software/ComputeLibrary/issues/402 if you still encounter any problems. The API will change in the next release.

Thanks

Issue #323, opened 2017-12-30
DepthwiseConvolutionLayer does not use optimised 3x3 convolutions despite opt3x3=true


psyhtest:
While profiling the MobileNet graph example using [dvdt-prof](https://github.com/dividiti/dvdt-prof) under [Collective Knowledge](https://github.com/ctuning/ck), I've noticed that despite `DepthwiseConvolutionLayer` being initialised as follows:
```
    SubGraph    sg;
    sg << DepthwiseConvolutionLayer(
           3U, 3U,
           get_weights_accessor(data_path, total_path + "depthwise_depthwise_weights.npy"),
           std::unique_ptr<arm_compute::graph::ITensorAccessor>(nullptr),
           dwc_pad_stride_info,
           true)
```
(where the last argument `opt3x3=true` [means](https://github.com/ARM-software/ComputeLibrary/blob/master/arm_compute/graph/nodes/DepthwiseConvolutionLayer.h#L52) using `DepthwiseConvolutionLayer3x3`), the "direct" `depthwise_convolution_3x3` kernel is not used. Instead, a significantly less optimised sequence of kernels is called including `depthwise_im2col`, `depthwise_weights_reshape`, `depthwise_vector_to_tensor` and `gemm_mv`. Could you confirm please?

AnthonyARM:
Hi Anton,
We'll have a look but it seems to be the same issue as #322

psyhtest:
Thanks @AnthonyARM! Yes, it's similar to #322 "semantically", although quite different from it "syntactically".

mohamedadaly:
Was this issue resolved? I am trying to compare ACL's graph_mobilenet to Tensorflow Lite, and I can see that ACL is much slower. For an image of 224x224 and using 4 threads: ACL takes about 1.3 secs white Tensorflow Lite takes about 0.5 sec. This is on an ARMv7 processor.

Do you have an intuition why ACL is much slower?

AnthonyARM:
What platform is this on ?
How did you time the network ? How many iterations did you do ?

mohamedadaly:
- I am running on a custom board based on Qualcomm's APQ8009 SoC.
- Timed the network like so:
```C
        // Run graph
        for (int i = 0; i < 5; ++i) {
            auto t1 = std::chrono::steady_clock::now();
            graph.run();
            auto t2 = std::chrono::steady_clock::now();
[...]

Issue #322, opened 2017-12-29
ConvolutionMethodHint ignored in SubGraph's?


psyhtest:
I'm using [dvdt-prof](https://github.com/dividiti/dvdt-prof) to profile OpenCL graph examples under [Collective Knowledge](https://github.com/ctuning/ck).

To my surprise, for the SqueezeNet example, direct 1x1 convolution is only used 9 times for the `squeeze1x1` layers, while GEMM-based convolution is used 16 times for the `expand1x1` and `expand3x3` layers (and once for the 7x7 `conv1` layer). In other words, it appears that `convolution_hint = ConvolutionMethodHint::DIRECT` only gets applied to the convolutions directly added to the graph in `main_graph_squeezenet()`, but not to the convolutions indirectly added to the graph via `get_expand_fire_node()`.

Could you confirm please?

psyhtest:
Unfortunately, trying to force the hint for the subgraphs:
```
diff --git a/examples/graph_squeezenet.cpp b/examples/graph_squeezenet.cpp
index c8c411aa..a79bd80d 100644
--- a/examples/graph_squeezenet.cpp
+++ b/examples/graph_squeezenet.cpp
@@ -46,6 +46,7 @@ BranchLayer get_expand_fire_node(const std::string &data_path, std::string &&par
             get_weights_accessor(data_path, total_path + "expand1x1_w.npy"),
             get_weights_accessor(data_path, total_path + "expand1x1_b.npy"),
             PadStrideInfo(1, 1, 0, 0))
+        << ConvolutionMethodHint(ConvolutionMethodHint::DIRECT)
         << ActivationLayer(ActivationLayerInfo(ActivationLayerInfo::ActivationFunction::RELU));

     SubGraph i_b;
```
results in:
```
In file included from ./arm_compute/graph/nodes/BranchLayer.h:30:0,
                 from ./arm_compute/graph/Nodes.h:29,
                 from examples/graph_squeezenet.cpp:25:
./arm_compute/graph/SubGraph.h: In instantiation of ‘arm_compute::graph::SubGraph& arm_compute::graph::operator<<(arm_compute::graph::SubGraph&, Node) [with Node = arm_compute::graph::ConvolutionMethodHint]’:
examples/graph_squeezenet.cpp:49:63:   required from here
[...]

PR #432, opened 2018-04-17
ReQuEST direct convolution kernel

psyhtest:
This PR provides a new direct convolution kernel which we used for [dividiti's submission](https://github.com/dividiti/ck-request-asplos18-mobilenets-armcl-opencl) to the [1st ReQuEST competition at ASPLOS'18](http://cknowledge.org/request-cfp-asplos2018.html).

As I will prove shortly, this kernel is faster than GEMM-based convolution for 15 out of 16 MobileNets models (i.e. all except the baseline model with the width multiplier of 1.0 and the resolution of `224x224`).


psyhtest:
Here's how to read the graph:
- The colour of a marker denotes the version of ArmCL: "17.12", "18.01", "18.03", "dv/dt" (this PR from a fork of "18.03").
- The size of a marker is proportional to the input resolution: 224 (the largest), 192, 160, 128 (the smallest).
- The shape of a marker denotes both the channel multiplier and the convolution method:
  - Direct convolution: 1.00 - pentagon, 0.75 - square, 0.50 - triangle-up, 0.25 - circle.
  - GEMM-based convolution: 1.00 - star, 0.75 - diamond, 0.50 - triangle-down, 0.25 - octagon.

Roughly speaking, a shape has **N** corners for the channel multiplier of **(N-1)*0.25**:
- 5 corners for the channel multiplier of 1.0 (pentagon or star);
- 4 corners for the channel multiplier of 0.75 (square or diamond);
- 3 corners for the channel multiplier of 0.50 (triangle-up or triangle-down); 
- "no corners" for the channel multiplier of 0.25 (circle or octagon).

psyhtest:
The plot bounds are chosen so that all the 32 points for ArmCL 18.03 are shown (32 = 4 input resolutions times 4 channel multiplies times 2 convolution methods). Only a few points are shown for ArmCL 17.12 for two reasons:
1. dividiti's optimisations for direct convolution kernels were mostly disabled in 17.12 due to bugs (https://github.com/ARM-software/ComputeLibrary/issues/322, https://github.com/ARM-software/ComputeLibrary/issues/323), which were fixed in 18.01. Therefore, the most accurate points were simply too slow with 17.12 (> 100 ms).
2. For some points, accuracy dropped practically to ~10%. Presumably, this was due to some other bug in 17.12 which also got fixed in 18.01.

psyhtest:
This graph plots Top 1 accuracy on all the 50,000 images of the ImageNet validation set (Y, %) against the image classification time (X, ms). Due to the long evaluation time across all the 16 models, accuracy was only evaluated using ArmCL 18.03, and then reused for the other library variants (therefore, no accuracy degradation for ArmCL 17.12 described above is visible).

accuracy_top1_ 50000 dv_dt 18_03 18_01 17_12

psyhtest:
In summary, using direct convolution is faster than GEMM-based convolution for 15 MobileNets models out of 16 (i.e., all except the baseline model with the channel multiplier of 1.00 and input resolution of 224). Also, it's practically possible to exceed AlexNet-level accuracy (0.59 > 0.57) in only about 17 ms per image.

AnthonyARM:
FYI: We haven't forgotten about you but we're a bit tight for the merge window for 18.05, so I moved this PR to 18.08

psyhtest:
@AnthonyARM I am looking at resolving the conflict. But I don't see the code for setting the local work size hint (`_lws_hint`) any more. Do you completely rely on online tuning now please?

psyhtest:
OK, I've found it [here](https://github.com/ARM-software/ComputeLibrary/blob/master/src/runtime/CL/tuners/BifrostTuner.cpp).

Now, we have heuristics to submit for setting up `_lws_hint` for a couple of other kernels. Would you suggest to place them similarly to this file?

AnthonyARM:
@psyhtest Yes, that's correct, we're going to move all the heuristics to that file.

We've also got a ticket to look at this PR which is getting near the top of our backlog now.

psyhtest:
Thanks @AnthonyARM. We've clearly been unlucky with direct convolution being moved there first :). Perhaps it will be easier to merge it on your side then. The heuristics are still the same.

GeorgeARM:
I tried to merge this pull request and found out that is >1.5x slower in both dynamic and static tuning against the gemm approach for the two mobilenet graph examples that we support in benchmarks. Can you rebase and check against 18.05?

psyhtest:
Hi @GeorgeARM, As the above graph shows, this kernel is indeed slower for the baseline version (although not > 1.5x slower). I'll check against v18.05 in 1-2 weeks.

I suspect the reason it is slower relates to your second comment. This **used** to be correct in v17.12-v18.03 (https://github.com/ARM-software/ComputeLibrary/blob/c3f34a43ffb5d52ee4a4e9f7b1bf4c6c002aeebb/src/core/CL/kernels/CLDirectConvolutionLayerKernel.cpp#L331), but is now broken in v18.05 (https://github.com/ARM-software/ComputeLibrary/blob/b3a371bc429d2ba45e56baaf239d8200c2662a74/src/runtime/CL/tuners/BifrostTuner.cpp#L68).

The trailing underscore there was intentional - to signal that it was "sort of product of input dimensions" but not quite (please read the comment above).

GeorgeARM:
Closing this pull request as no performance benefit has been noted in latest ACL.
Reopen if needed. Thank you

psyhtest:
On your request, I have provided further evidence on the performance our ReQuEST kernel :). Unfortunately, I do not seem to have permissions to re-open this pull request? ("request" again!:))

Issue #405, opened 2018-04-04
build library but always print "Cannot compile NEON for x86"


lebronze:
I followed the Compute Library  18.03 documents : " build for Android"  to configure everything in my computer, and i successfully configured everything.

Then i tried to build the library:
To cross-compile the library in debug mode, with NEON only support, for Android 32bit:
CXX=clang++ CC=clang scons Werror=1 -j8 debug=1 neon=1 opencl=0 os=android arch=armv7a

but error comes:  
    File "/home/qb/ComputeLibrary-master/SConstruct", line 74
    print "Cannot compile NEON for x86"
    SyntaxError: invalid syntax

I didn't use x86,  my commond is "arch=v7a"   !!
The i tried to make "neon=0", "arc=arm64-v8a".  However, the same mistake comes.

AnthonyARM:
As you can see inside the [SConstruct file](https://github.com/ARM-software/ComputeLibrary/blob/master/SConstruct#L73) this error will only be printed if you're trying to compile with `neon=1` and `arch=x86_32` or `arch=x86_64`

Can you try to add a `print env['arch']` on line 72 to check which architecture is selected ?

lebronze:
thank you very much!
 i tried print env['arch'], but error comes :

  File "/home/qubo/ComputeLibrary-master/SConstruct", line 72
    print env['arch']
           ^
SyntaxError: invalid syntax

is it possible that the version of python cause the mistake?

AnthonyARM:
Possibly, I think SCons only works with python2, in `/usr/bin/scons` I've got:

```python
if sys.version_info >= (3,0,0):
    msg = "scons: *** SCons version %s does not run under Python version %s.\n\
Python 3 is not yet supported.\n"
    sys.stderr.write(msg % (__version__, sys.version.split()[0]))
    sys.exit(1)
```
So in theory it shouldn't run at all if all you've got on your system is python 3.

lebronze:
Em, i check my computer and find the version of python in my computer is python 2.7.
I don't konw why this error: SyntaxError: invalid syntax  comes.

lebronze:
oh， i solved this problem by  change:   print "Cannot compile NEON for x86"   into   print("Cannot compile NEON for x86")  in the file SConstruct.
It's about the usage of print in Python.

AnthonyARM:
No, it's definitely something to do with the version of python you're using
In my case if I add
```python
import sys
print sys.version
```
Then I get:
```
[...]

PR #454, opened 2018-06-01
Fix build

ggardet:
This PR fixes build with new GCC and with Python 3.

This will fix https://github.com/ARM-software/ComputeLibrary/issues/405


AnthonyARM:
Looks good, thanks for fixing these issues

PR #460, opened 2018-06-10
Deconvolution fixes

wr0112358:
Please consider pulling these patches for NEON and CL deconvolution layers.
The first one is the same as #439 but for NEON.
The others improve FP16 support. I tested the FP16 deconvolution layer with a Resnet-Linknet.


mpekatsoula:
The changes will be available in the next release.

Thanks

PR #518, opened 2018-08-06
Bugfix of subtensor validation

llFreetimell:
Shape of parent tensor should be compared with shape of subtensor, not
with itself.

Signed-off-by: llFreetimell <seok9311@naver.com>


GeorgeARM:
Cherry picked. Will be part of the next release.

mdigiorgio:
This change is now part of the v18.08 release. Thanks for your contribution.

PR #525, opened 2018-08-23
Removed redundant const qualifier on cast

Flast:
GCC (>=8) yields warning w/ -Wignored-qualifers (enabled by -Wextra) on such usage.


mdigiorgio:
This change is now part of the v18.08 release. Thanks for your contribution.

PR #527, opened 2018-08-24
Enable QASYMM8 in LOGISTIC of CLActivationLayerKernel.

jyoungyun:
This commit enables QASYMM8 in the LOGISTIC activation function.
The datatype of inputs and output must all be QASYMM8.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>


gmiodice:
Hi @jyoungyun,

many thanks for your pull request about "logistic with QASYMM8". 

In the coming release 18.08 we have introduced this operator for the activation layer accelerated with floating point (this means we convert the value from QASYMM8 to Float, we compute the logistic and we convert back to QASYMM8).
The reason why we implemented in this manner is because we found fewer arithmetic instructions with floating point rather than computing the logistic with integer.

Many thanks for proposing this patch anyway and for using ACL.

jyoungyun:
@gmiodice Thank you for comments. I understood what direction your team was working in. After 18.08 is released, I will apply your changes in our project and if there is some issue, I will make a different issue. I close my issue.

PR #554, opened 2018-10-02
check bias data type only if it is not null

we1mingzha0:
line 69 will check if bias data type matches


GeorgeARM:
Fixed on [master](https://git.mlplatform.org/ml/ComputeLibrary.git/tree/src/runtime/NEON/functions/NEDeconvolutionLayer.cpp?h=refs/heads/master#n71).

PR #1181, opened 2025-09-16
perf: Add FP16 GEMM MMUL Reshaped Only Rhs Kernel

ArmDude:
This patch introduces a GEMM routine that is optimized for Arm(R) Mali(TM)-G1

Resolves: [COMPMID-8311], [COMPMID-8312]

Change-Id: I84e685f0314da9af1c3fbb50d83e68b355727770

### DIFF
--- a/arm_compute/core/CL/cl_definitions.h
+++ b/arm_compute/core/CL/cl_definitions.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2025 Arm Limited.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef ACL_ARM_COMPUTE_CORE_CL_CL_DEFINITIONS_H
+#define ACL_ARM_COMPUTE_CORE_CL_CL_DEFINITIONS_H
+
+#include "include/CL/opencl.hpp"
+
+#define CL_DEVICE_MATRIX_MULTIPLY_FP16_WITH_FP16_ACCUMULATORS_ARM (1ULL << 0)
+#define CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM                0x41F4
+
+namespace cl
+{
+namespace detail
+{
+#ifdef CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM
+CL_HPP_DECLARE_PARAM_TRAITS_(cl_device_info, CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM, cl_ulong)
+#endif // CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM
+} // namespace detail
+} // namespace cl
+#endif // ACL_ARM_COMPUTE_CORE_CL_CL_DEFINITIONS_H
--- a/src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
+++ b/src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023 Arm Limited.
+ * Copyright (c) 2022-2023, 2025 Arm Limited.
  *
  * SPDX-License-Identifier: MIT
  *
@@ -24,6 +24,8 @@
 #include "activation_float_helpers.h"
 #include "helpers.h"
 #include "tile_helpers.h"
+
+#define SHIFT_BACK(IDX, N0, PARTIAL_N0) (max((int)((IDX) * N0 - (N0 - PARTIAL_N0) % N0), 0))
 
 #if defined(GEMM_MM_RESHAPED_ONLY_RHS_NT_MMUL)
 /** This OpenCL kernel computes the matrix multiplication between 2 matrices using the MMUL extension:
@@ -219,7 +221,7 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul(
 
 #ifndef UNIT_BETA
     T_SCALE_CONSTANT(DATA_TYPE, 1, N0, bias0, (DATA_TYPE)BETA, bias0);
-#endif // UNIT_BIAS
+#endif // UNIT_BETA
 
     // c = c + bias[broadcasted]
     T_ELTWISE_BROADCAST_X(V_ADD, DATA_TYPE, M0, N0, c, bias0, c);
@@ -252,7 +254,7 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul(
 
 #ifndef UNIT_BETA
     T_SCALE_CONSTANT(DATA_TYPE, M0, N0, bias0, (DATA_TYPE)BETA, bias0);
-#endif // UNIT_BIAS
+#endif // UNIT_BETA
 
     // c = c + bias
     T_ADD(DATA_TYPE, M0, N0, c, bias0, c);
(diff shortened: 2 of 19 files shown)


Dongsung-arm:
Looks fine to me

Dongsung-arm (approved):

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,10 +90,21 @@ Status validate_arguments(const ITensorInfo       *src0,
...
-    const unsigned int n = gemm_info.n;
-    const unsigned int k = gemm_info.k;
-
+    const unsigned int m       = gemm_info.m;
+    const unsigned int n       = gemm_info.n;
+    const unsigned int k       = gemm_info.k;
+    const bool         is_fp16 = (src0->data_type() == DataType::F16);
+
+    // These error messages are for FP16 acc.
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && (n < rhs_info.n0 * mmul_n0), "N must be greater that N0 * MMUL_N0");
```
gunes-arm: typo: greater than

gunes-arm: First suggestion: Put the fp16 related validations into a `if(is_fp16)` block and remove `is_fp16` from every check.

Also, the message should be specific to fp16 kernel, e.g.
"K must be multiple of 4 in fp16 mmul kernel".

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,10 +90,21 @@ Status validate_arguments(const ITensorInfo       *src0,
...
     ARM_COMPUTE_RETURN_ON_ERROR(gemm::validate_image2d_support_on_rhs(*src1, rhs_info));
 
-    const unsigned int m = gemm_info.m;
-    const unsigned int n = gemm_info.n;
-    const unsigned int k = gemm_info.k;
-
+    const unsigned int m       = gemm_info.m;
+    const unsigned int n       = gemm_info.n;
+    const unsigned int k       = gemm_info.k;
+    const bool         is_fp16 = (src0->data_type() == DataType::F16);
```
gunes-arm, line 96: We check all these validations for fp16, but at this point, we do not know whether we'll be using fp16 mmul kernel. So, we might be validating against all these for the fp32 mmul kernel running on fp16 input, which is wrong, right?

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,10 +90,21 @@ Status validate_arguments(const ITensorInfo       *src0,
...
+
+    // These error messages are for FP16 acc.
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && (n < rhs_info.n0 * mmul_n0), "N must be greater that N0 * MMUL_N0");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && ((k % 4) != 0), "K must be multiple of 4");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && (m < 4), "M must be greater or equal to 4");
```
gunes-arm: typo: greater than or equal to

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,10 +90,21 @@ Status validate_arguments(const ITensorInfo       *src0,
...
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && (lhs_info.k0 != 1), "k0 must be equal to 1");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 &&
+                                        (rhs_info.n0 != 2 && rhs_info.n0 != 4 && rhs_info.n0 != 8 && rhs_info.n0 != 16),
+                                    "Only 2,4,8,16, and 16 are supported for n0");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(is_fp16 && (m < lhs_info.m0 * mmul_m0),
+                                    "M must be greater or equal to m0 * mmul_m0");
```
gunes-arm: typo: greater than or ...

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -207,11 +219,10 @@ std::pair<Status, Window> validate_and_configure_window(ITensorInfo
 
     // Make n_div_n0 and m_div_m0 multiple of mmul_n0 and mmul_k0 respectively
```
gunes-arm: to be consistent with below, mmul_k0 should be changed to mmul_m0 as well

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
```diff
@@ -125,6 +125,42 @@ bool is_mmul_kernel_preferred(const unsigned int m,
...
+ *
+ * @param[in] configs List of best configurations for a limited number of GeMM shapes
+ * @param[in] m       Number of rows of the LHS matrix
+ * @param[in] n       Number of columns of the RHS matrix
+ * @param[in] k       Number of columns of the LHS matrix, rows of the RHS matrix
+ * @param[in] b       Batch size
+ *
+ * @return @ref GEMMLHSMatrixInfo and @ref GEMMRHSMatrixInfo
+ */
+bool is_mmul_kernel_preferred_fp16_acc(const unsigned int m,
```
gunes-arm: same function declared twice

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
```diff
@@ -125,6 +125,42 @@ bool is_mmul_kernel_preferred(const unsigned int m,
                               unsigned int      &best_m0,
                               unsigned int      &best_n0);
 
+/** Find the preferred configurations for the LHS and RHS tensor using the GeMMConfigsMatrix provided by the user
```
gunes-arm: what is GeMMConfigsMatrix?

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
```diff
@@ -125,6 +125,42 @@ bool is_mmul_kernel_preferred(const unsigned int m,
...
+ *
+ * @param[in] configs List of best configurations for a limited number of GeMM shapes
+ * @param[in] m       Number of rows of the LHS matrix
+ * @param[in] n       Number of columns of the RHS matrix
+ * @param[in] k       Number of columns of the LHS matrix, rows of the RHS matrix
+ * @param[in] b       Batch size
+ *
+ * @return @ref GEMMLHSMatrixInfo and @ref GEMMRHSMatrixInfo
```
gunes-arm: return type and argument list are not accurate

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -132,15 +139,15 @@ FIXTURE_DATA_TEST_CASE(RunSmall, CLGEMMMatrixMultiplyReshapedOnlyRhsMMULFixture<
 
 TEST_SUITE_END() // FP32
 
-TEST_SUITE(FP16)
+TEST_SUITE(MMUL_FP16)
```
gunes-arm: The fp32 mmul kernel still supports Fp16 although the heuristics don't choose it (same goes for block sizes), therefore we shouldn't remove this test.

Also, we should replace the `combine(combine(...` patterns with a single combine, i.e. `combine(...)` wherever we touch.

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -66,13 +66,17 @@ const auto beta_values = framework::dataset::make("beta", {0.0f, -0.75f} );
 
 /** M values to test */
 const auto m_values = framework::dataset::make("M", {49});
+const auto m_values_fp16 = framework::dataset::make("M", {131});
 
 /** N values to test */
-const auto n_values = framework::dataset::make("N", {257});
+const auto n_values              = framework::dataset::make("N", {257});
+const auto n_values_fp16         = framework::dataset::make("N", {79});
```
gunes-arm: I think one of the mistakes we did was to test on a single shape. While we test different block sizes on a single shape, we should also test on small shapes and a subset of block sizes. (While keeping in mind the test time required ofc.)

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -81,20 +85,23 @@ const auto b_values = framework::dataset::make("batch_size", {1, 2});
...
+const auto n0_values_precommit              = framework::dataset::make("N0", { 4, 8 });
+const auto n0_values_precommit_fp16         = framework::dataset::make("N0", { 2, 4, 8, 16 });
+const auto n0_values_precommit_texture_fp16 = framework::dataset::make("N0", { 4, 8 });
 
 /** K0 values to test - Precommit */
 const auto k0_values_precommit = framework::dataset::make("K0", { 1 });
 
 /** Broadcast bias from vector to matrix */
-const auto broadcast_bias_values = framework::dataset::make("broadcast_bias", { false, true } );
+const auto broadcast_bias_values = framework::dataset::make("broadcast_bias", { false } );
```
gunes-arm: This change reduces fp32 tests as well

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -160,7 +167,38 @@ FIXTURE_DATA_TEST_CASE(RunSmall, CLGEMMMatrixMultiplyReshapedOnlyRhsMMULFixture<
         framework::ARM_COMPUTE_PRINT_INFO();
     }
 }
-TEST_SUITE_END() // FP16
+
+TEST_SUITE(ExportToCLImage)
```
gunes-arm: If there will be any, they should be under the TEST_SUITE(ExportToCLImage) below

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -160,7 +167,38 @@ FIXTURE_DATA_TEST_CASE(RunSmall, CLGEMMMatrixMultiplyReshapedOnlyRhsMMULFixture<
...
+                                                                   k0_values_precommit),
+                                                                   framework::dataset::make("ExportToCLImage", true)),
+                                                                   framework::dataset::make("DataType", DataType::F16)),
+                                                                   a_values),
+                                                                   beta_values),
+                                                                   broadcast_bias_values),
+                                                                   act_values))
+{
+    // Validate output
+    if(validate_result)
```
gunes-arm, line 157: I think we've done this the wrong way 4 years ago, and the culprit is me :) We shouldn't do the same mistake here. If you look at GEMMFixture.h and the relevant fixture class, this is set to false when the gemm or reshape does not validate to true. So, if we provide a faulty configuration to test, it'll skip the test. This is definitely not the right thing to do.

I hope the context is clear. Let me know if it's not. Now, what do we need to do?:

We change GEMMMatrixMultiplyReshapedOnlyRhsMMULValidationFixture so that validate_result is set to false only if the hardware features relevant for the test in question are not supported.

i.e. We have fp32 and fp16 tests using mmul with fp32 accumulators and we have fp16 tests using mmul with fp16 accumulators.

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
ArmDude: How about this?

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
ArmDude: How about this? (same change as for the comment before it)

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
ArmDude: Typo, this documentation comment was copy pasted from the function at the bottom of this file find_lhs_rhs_info()

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
ArmDude: How about this? The documentation comments are the same as FP32 version with the only difference being the first line specifying the data type for both function?

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: How about this? These are values from a number of models of interest, given the combinatorial expansion, I think 3 is a reasonable number of values for each of the datasets. I've tried to pick shapes on the smaller side to keep runtime reasonable too.

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: How's this?

arm_compute/core/CL/cl_definitions.h
```diff
@@ -0,0 +1,42 @@
...
+#include "include/CL/opencl.hpp"
+
+#define CL_DEVICE_MATRIX_MULTIPLY_FP16_WITH_FP16_ACCUMULATORS_ARM (1ULL << 0)
+#define CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM                0x41F4
+
+namespace cl
+{
+namespace detail
+{
+#ifdef CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM
```
dwildmark, line 37: This will always be defined, why does it need to be ifdef'd?

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: I've made a change to GEMMFixture.h to the setting of this value, now making it correspond to whether the target hardware supports arm_matrix_multiply.

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: How about this?

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
```diff
@@ -292,6 +294,310 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul(
...
+    }
+    else
+    {
+        VLOAD_PARTIAL(N0, N0_LEFTOVER)
+        (bias0[0].v, 0, (DATA_TYPE *)(bia_ptr + bia_offset_first_element_in_bytes));
+    }
+
+#ifndef UNIT_BETA
+    T_SCALE_CONSTANT(DATA_TYPE, 1, N0, bias0, (DATA_TYPE)BETA, bias0);
+#endif // UNIT_BIAS
```
gunes-arm: UNIT_BETA (there is another one below)

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
```diff
@@ -292,6 +294,310 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul(
...
+        {
+            LOOP_UNROLLING(int, m0, 0, 1, M0_LEFTOVER,
+            {
+                VSTORE_PARTIAL(N0, N0_LEFTOVER)
+                (c[m0].v, 0, (__global DATA_TYPE *)(dst_ptr + dst_offset_first_element_in_bytes + m0 * dst_stride_y));
+            })
+        }
+    }
+
+#undef RHS_BLOCK_SIZE
```
gunes-arm: where are these defined? (Same in texture kernel)

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
```diff
@@ -554,3 +860,306 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul_texture(
...
+ *       The activation function is performed after the bias addition
+ *
+ * @param[in]  lhs_ptr                           Pointer to the LHS tensor. Supported data types: F16
+ * @param[in]  lhs_stride_y                      Stride of the LHS tensor in Y dimension (in bytes)
+ * @param[in]  lhs_stride_z                      Stride of the LHS tensor in Z dimension (in bytes)
+ * @param[in]  lhs_w                             The size of the width dimension of the LHS tensor
+ * @param[in]  lhs_h                             The size of the height dimension of the LHS tensor
+ * @param[in]  lhs_n                             The size of the depth dimension of the LHS tensor
+ * @param[in]  lhs_offset_first_element_in_bytes The offset of the first element in the LHS tensor
+ * @param[in]  rhs_ptr                           Pointer to the RHS reshaped tensor. Supported data type: same as @p lhs_ptr
```
gunes-arm, line 879: There is also `rhs_img` (check #define TENSOR3D_T_IMAGE)

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,9 +90,28 @@ Status validate_arguments(const ITensorInfo       *src0,
...
+        {
+            // These error messages are for FP16 acc.
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG((n < rhs_info.n0 * mmul_n0),
+                                            "N must be greater than N0 * MMUL_N0 in the FP16 MMUL Kernel");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(((k % 4) != 0), "K must be multiple of 4 in the FP16 MMUL Kernel");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG((m < 4), "M must be greater than or equal to 4 in the FP16 MMUL Kernel");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG((lhs_info.k0 != 1), "k0 must be equal to 1 in the FP16 MMUL Kernel");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(
+                (rhs_info.n0 != 2 && rhs_info.n0 != 4 && rhs_info.n0 != 8 && rhs_info.n0 != 16),
+                "Only 2,4,8,16, and 16 are supported for n0 in the FP16 MMUL Kernel");
```
gunes-arm: 16 is written twice

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
```diff
@@ -89,9 +90,28 @@ Status validate_arguments(const ITensorInfo       *src0,
...
+    const unsigned int k       = gemm_info.k;
+    const bool         is_fp16 = (src0->data_type() == DataType::F16);
+
+    if (is_fp16)
+    {
+        if (arm_matrix_multiply_fp16_supported(CLKernelLibrary::get().get_device()))
+        {
+            // These error messages are for FP16 acc.
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG((n < rhs_info.n0 * mmul_n0),
+                                            "N must be greater than N0 * MMUL_N0 in the FP16 MMUL Kernel");
```
gunes-arm, line 104: looks like this should be "greater than or equal to"

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
```diff
@@ -105,7 +105,7 @@ void update_padding_for_cl_image(ITensorInfo *tensor);
  */
 Status validate_image2d_support_on_rhs(const ITensorInfo &tensor_reshaped_info, const GEMMRHSMatrixInfo &rhs_info);
 
-/** Determine if the MMUL kernels should be preferred
+/** Determine if the FP32 MMUL kernels should be preferred
```
gunes-arm, line 108: We should rename this function as `is_mmul_kernel_preferred_fp32_acc`

gunes-arm:
I suggest to change the title to: "Add FP16 GEMM MMUL Reshaped Only Rhs Kernel"

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
```diff
@@ -66,13 +66,17 @@ const auto beta_values = framework::dataset::make("beta", {0.0f, -0.75f} );
 
 /** M values to test */
 const auto m_values = framework::dataset::make("M", {49});
+const auto m_values_fp16 = framework::dataset::make("M", {131});
 
 /** N values to test */
-const auto n_values = framework::dataset::make("N", {257});
+const auto n_values              = framework::dataset::make("N", {257, 448, 384});
```
gunes-arm: I think we can choose the second and third shapes smaller. These are too big. Leave the first one though.

gunes-arm (changes requested):

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
ArmDude: Good spot

src/gpu/cl/kernels/ClGemmMatrixMultiplyReshapedOnlyRhsMMULKernel.cpp
ArmDude: Good spot

src/gpu/cl/kernels/gemm/ClGemmHelpers.h
ArmDude: Indeed, to distinguish it from FP16

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: How are these?

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
ArmDude: Done

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
ArmDude: Accidentally copied over from gemm.cl

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
ArmDude: How's this?

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
```diff
@@ -285,13 +287,309 @@ __kernel void gemm_mm_reshaped_only_rhs_nt_mmul(
...
+#if(N0 != 2 && N0 != 4 && N0 != 8 && N0 != 16)
+#error "N0 can only be 2,4,8,16"
+#endif
+
+#if(K0 != 1)
+#error "K0 can only be 1"
+#endif
+
+#define MMUL_BLOCK_SIZE (MMUL_N0 * MMUL_M0)
+#define VEC_SIZE 2 // For fp16 types input to mmul instruction is a length 2 vector
```
gunes-arm: Where is this used? Implicitly inside a macro, or not at all?

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
gunes-arm: I think they are still big. We'll battle with tolerance issues if we set these shapes too large (also considering the test time required). I just want some additional coverage, that's it. It doesn't have to be big. For example, I don't think we need M = 131 case.

const auto n_values_texture_fp16 = framework::dataset::make("N", {128, 192, 160});
192 & 160 are big. Try to put much smaller values that this kernel can run.

src/core/CL/cl_kernels/common/gemm_reshaped_only_rhs_mmul.cl
ArmDude: Ah good spot, not used

tests/validation/CL/GEMMMatrixMultiplyReshapedOnlyRhsMMUL.cpp
ArmDude: How's this?

PR #1184, opened 2025-09-19
feat: enable F32 output in CpuGemmConv2d

morgolock:
- Updated convolution reference to branch epilogue:
  * TO=float: int32 to float dequant (acc * sA * sB + bias_f32)
  * TO!=float: usual quantize_down_scale_by_fixedpoint with int32 bias
- Changed fixture to use F32 bias tensor for Q->F32 runs (instead of S32), matching arm_gemm dequant epilogue which only supports float bias.
- Added explicit template instantiations for convolution_layer with TBias=float, TO=float to fix linker errors in validation.
- Disabled activation in arm_gemm dequant path: offsets are applied afterwards by CpuGemmLowpOffsetContributionKernel, so activation must run there to see the correct final accumulator.

This aligns target and reference for quantized to F32 convolution tests and prevents premature clamping before offset contributions.

Change-Id: I6fffc98dc0798542a2702e6a593b850c16561e3b


tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -386,6 +409,14 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
             conv.run();
         }
 
+        std::cout << "src\n";
```
wajahat-abbas: Are these statements intentional?

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -534,19 +564,22 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
...
             src = reference::pad_layer<T>(src, pre_pad_layer, PixelValue(0), PaddingMode::CONSTANT);
         }
 
-        return (act_info.enabled()) ? reference::activation_layer<T>(reference::convolution_layer<T>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info),
+        auto res=  (act_info.enabled()) ? reference::activation_layer<TO>(reference::convolution_layer<T,TW,TBias,TO>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info),
                                                                      act_info) :
-               reference::convolution_layer<T>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info);
+               reference::convolution_layer<T,TW,TBias,TO>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info);
+
+        print_simpletensor(res,"ref res");
```
wajahat-abbas: Is this intentional?

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Done. Removed now.

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Done. Thanks, removed now.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -1397,6 +1401,21 @@ FIXTURE_DATA_TEST_CASE(RunSmall, NEGEMMConvolutionLayerForUpdatedStaticQuantInfo
     // Validate output
     validate(Accessor(_target), _reference, tolerance_qasymm8);
 }
+
+FIXTURE_DATA_TEST_CASE(RunSmallDequantizeF32, NEGEMMConvolutionLayerQuantizedF32OutputFixture<int8_t>, framework::DatasetMode::ALL, combine(combine(combine(combine(combine(datasets::SmallConvolutionLayerDataset(),
```
gunes-arm: single combine(..) is enough

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -534,19 +555,20 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
...
                                                                      act_info) :
-               reference::convolution_layer<T>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info);
+               reference::convolution_layer<T,TW,TBias,TO>(src, weights, bias, output_shape, info, dilation, num_groups, _dst_q_info);
+
+        return res;
     }
 
     TensorType       _target{};
-    SimpleTensor<T>  _reference{};
+    SimpleTensor<TO>  _reference{};
```
gunes-arm: space alignment

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -353,7 +375,10 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
         ARM_COMPUTE_ASSERT(bias.info()->is_resizable());
         ARM_COMPUTE_ASSERT(dst.info()->is_resizable());
         // Test "add padding after configure" behavior. This behavior should not affect the correctness
-        add_padding_x({ &src, &bias, &dst }, _data_layout);
+        if( !(T_is_q && TO_is_f32))
```
gunes-arm: I believe we should enable this now

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -471,7 +495,10 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
...
+        {
+            add_padding_x({ &src, &bias, &dst }, _data_layout);
```
gunes-arm: similarly

tests/datasets/SmallConvolutionLayerDataset.h
```diff
@@ -21,8 +21,8 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-#ifndef ARM_COMPUTE_TEST_SMALL_CONVOLUTION_LAYER_DATASET
-#define ARM_COMPUTE_TEST_SMALL_CONVOLUTION_LAYER_DATASET
+#ifndef ACL_TESTS_DATASETS_SMALLCONVOLUTIONLAYERDATASET_H
```
gunes-arm: I suppose we don't need to change this file

src/cpu/kernels/gemmlowp/generic/neon/impl.h
```diff
@@ -216,7 +217,7 @@ void neon_run_offset_contribution_float(const Window  &window,
                 const int    batch_id = id.z() / depth_input;
                 const size_t batch_offset_col =
                     batch_id *
-                    sum_col_stride_y; // Value to offset vector_sum_col_ptr to allow for iteration of y values in tensor
+                    sum_col_stride_w; // Value to offset vector_sum_col_ptr to allow for iteration of y values in tensor
```
gunes-arm: Does the comment still hold? i.e. iteration of y values

src/cpu/kernels/gemmlowp/generic/neon/impl.h
```diff
@@ -96,7 +97,7 @@ void neon_run_offset_contribution_float(const Window  &window,
             [&](const Coordinates &id)
             {
                 const int    batch_id         = id.z() / depth_input;
-                const size_t batch_offset_col = batch_id * sum_col_stride_y;
+                const size_t batch_offset_col = batch_id * sum_col_stride_w;
```
gunes-arm, line 100: I believe this file was added for Fp16 dequantization. How was it working for Fp16? Are we not testing batched inputs?

src/cpu/operators/CpuGemmConv2d.cpp
```diff
@@ -504,9 +537,11 @@ void CpuGemmConv2d::configure(const ITensorInfo         *src,
     }
 
     const unsigned int mat_weights_cols = weights->dimension(idx_kernels);
+    const bool         dequantize_f32   = is_data_type_quantized(data_type) && dst->data_type() == DataType::F32;
```
gunes-arm: Is is_data_type_quantized(.) check accurate? Or, are we only supporting Int8?

I think it's also confusing to add more than one meaning to `dequantize_f32` variable. We should probably have `is_input_int8` to denote Int8 inputs. 

How are we handling mixed signs? Do they validate to false? Can we add some validation() tests for this?

src/cpu/operators/CpuGemmConv2d.cpp
```diff
@@ -287,12 +287,29 @@ void CpuGemmConv2d::configure_mm(const ITensorInfo         *src,
...
         GEMMLowpOutputStageInfo output_info;
-        output_info.type                     = GEMMLowpOutputStageType::QUANTIZE_DOWN_FIXEDPOINT;
-        output_info.gemmlowp_offset          = uoqinfo.offset;
-        output_info.gemmlowp_min_bound       = min_activation;
-        output_info.gemmlowp_max_bound       = max_activation;
-        output_info.is_quantized_per_channel = (tmp_weights.data_type() == DataType::QSYMM8_PER_CHANNEL);
-        quantization::calculate_quantized_multipliers(iqinfo, wqinfo, oqinfo, output_info);
+
+        // F32 dequant path? (input quantized, output float)
+        const bool dequantize_f32 = (dst->data_type() == DataType::F32);
```
gunes-arm: We might also need to check input type. _is_quantized is true when we have UInt8 and Int16.

src/cpu/operators/CpuGemmConv2d.cpp
```diff
@@ -776,8 +818,9 @@ Status CpuGemmConv2d::validate(const ITensorInfo         *src,
         gemm_input_to_use = &im2col_reshaped_info;
     }
 
+    const bool dequantize_f32 = is_data_type_quantized(data_type) && dst->data_type() == DataType::F32;
```
gunes-arm: similarly for is_data_type_quantized

src/cpu/operators/CpuGemmLowpMatrixMultiplyCore.cpp
```diff
@@ -200,7 +206,7 @@ void CpuGemmLowpMatrixMultiplyCore::configure(
             case DataType::U8:
             case DataType::S8:
             {
-                if (is_data_type_quantized_asymmetric(a_to_use->data_type()) &&
+                if (dst->data_type() != DataType::F32 && is_data_type_quantized_asymmetric(a_to_use->data_type()) &&
```
gunes-arm: !dequantize_f32

src/cpu/operators/CpuGemmLowpMatrixMultiplyCore.cpp
```diff
@@ -488,6 +495,41 @@ Status CpuGemmLowpMatrixMultiplyCore::validate(const ITensorInfo *a,
         {
             ARM_COMPUTE_RETURN_ERROR_ON(a->dimension(1) != output->dimension(1));
         }
+
+        // Q -> F32 path, we add offsets in ACL so we need to validate:
+        // CpuGemmLowpMatrixAReductionKernel
+        // CpuGemmLowpMatrixAReductionKernel
+        if (a->data_type() == DataType::QASYMM8_SIGNED && output->data_type() == DataType::F32)
+        {
+            TensorInfo info_vector_sum_col{};
```
gunes-arm, line 540: I think we can move this logic to a function and reuse it in the place where you copied from.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -1363,6 +1363,10 @@ template <typename T>
 using NEGEMMConvolutionLayerForUpdatedStaticQuantInfoAfterConfigureFixture = ConvolutionValidationForUpdatedStaticQuantInfoAfterConfigureFixture<Tensor, Accessor, NEGEMMConvolutionLayer, T>;
 template <typename T>
 using NEGEMMConvolutionLayerQuantizedFixture = ConvolutionValidationQuantizedFixture<Tensor, Accessor, NEConvolutionLayer, T>;
+template <typename T>
+using NEGEMMConvolutionLayerQuantizedF32OutputFixture = ConvolutionValidationQuantizedFixture<Tensor, Accessor, NEConvolutionLayer, T,false,float>;
```
gunes-arm: We need to modify the public header files' comment headers to reflect the data type support as well as the operator list document.

Looks like there are two entry points. NEConvolutionLayer and NEGEMMConvolutionLayer. We need to be testing both.

Also, is it guaranteed to choose CpuGemmConv2d as the algorithm when this is triggered via NEConvolutionLayer? The get_convolution_method in CpuConv2d doesn't account for it. Or, do we have some configurations that can't be run for this data type configuration?

gunes-arm: We might also need to check if we accidentally return true in NEDeconv as it's using NEConv.

gunes-arm (changes requested):

src/cpu/kernels/gemmlowp/generic/neon/impl.h
morgolock: That's right. This patch is the first one to use `neon_run_offset_contribution_float` with batches

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Done in next patchset

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Renabled

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Enabled

src/cpu/operators/CpuGemmConv2d.cpp
morgolock: Done

src/cpu/operators/CpuGemmConv2d.cpp
morgolock: done

src/cpu/kernels/gemmlowp/generic/neon/impl.h
morgolock: Updated comment

src/cpu/operators/CpuGemmLowpMatrixMultiplyCore.cpp
morgolock: Moved to an inline function

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Added new test DequantFP32_SupportedTypes

src/cpu/operators/CpuConv2d.h
```diff
@@ -88,6 +88,7 @@ class CpuConv2d : public ICpuOperator
      * |QASYMM8        |QASYMM8_SIGNED     |S32    |QASYMM8        |
      * |QASYMM8        |QSYMM8_PER_CHANNEL |S32    |QASYMM8        |
      * |QASYMM8_SIGNED |QASYMM8_SIGNED     |S32    |QASYMM8_SIGNED |
+     * |QASYMM8_SIGNED |QASYMM8_SIGNED     |F32    |F32            |
```
gunes-arm, line 91: We also need to modify CpuGemmConv2d header

morgolock: Done

src/cpu/operators/CpuGemmLowpMatrixMultiplyCore.cpp
gunes-arm: I was referring to the code inside the conditional. Starting with TensorInfo and ending with if(b_offset_kernel_needed) code segment.

morgolock: Done in the latest patch.

tests/validation/NEON/ConvolutionLayer.cpp
gunes-arm: I think you've changed the one in QASYMM8 block :) This one is still with multiple combines.

tests/validation/NEON/ConvolutionLayer.cpp
gunes-arm: Thanks for adding the supported types. 

For the layers that we should test, i.e. NEConvolutionLayer and NEGEMMConvolutionalLayer, we should 
- Use `NEGEMMConvolutionLayer` in the declaration of `NEGEMMConvolutionLayerQuantizedF32OutputFixture`
- Create another fixture instantiation with `NEConvolutionLayer` and fixture name `NEConvolutionLayerQuantizedF32OutputFixture`. At the moment the fixture name and the layer name are mixed.
This way, we'll be able to test both layers.

I've also looked at the `get_convolution_method`. It's not guaranteed to choose CpuGemmConv2d for this configuration. Set channels to smth > 16, e.g. 22 in the first configuration in the `SmallConvolutionLayerDataset`; the test fails. This is because we're not validating correctly in `CpuGemmDirectConv2d`. I'll create a PR with a fix, and we can rebase on top of that. This brings us to my other observation, which is, we need to test on more shapes in the NIGHTLY suite. When I changed `SmallConvolutionLayerDataset` with `LargeConvolutionLayerDataset`, it again failed. We don't have enough coverage at the moment.

I've finally looked into `NEDeconvolutionLayer` which uses `NEConvolutionLayer` under the hood. It is fine as its validate() call is able to catch the Input/Dst data type configuration, unlike `CpuGemmDirectConv2d`.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -1397,6 +1473,21 @@ FIXTURE_DATA_TEST_CASE(RunSmall, NEGEMMConvolutionLayerForUpdatedStaticQuantInfo
...
+                                                            framework::dataset::make("QuantizationInfoIfActivationEnabled", { QuantizationInfo(2.f / 255.f, 10) })),
+                                                            ActivationFunctionsDataset))
+{
+    // Validate output
+    validate(Accessor(_target), _reference, rel_tolerance_f32, 0.f, float(abs_tolerance_f32));
+}
+
+
+
+
```
gunes-arm: We are inside `UpdateStaticQuantInfoAfterConfigure` suite. We should be outside of it.

gunes-arm (changes requested):

src/cpu/operators/CpuGemmLowpMatrixMultiplyCore.cpp
morgolock: done in latest patch

tests/datasets/SmallConvolutionLayerDataset.h
morgolock: Reverted in latest patch

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Fixed in next patch

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Moved the tests outside `UpdateStaticQuantInfoAfterConfigure`

src/cpu/kernels/gemmlowp/generic/neon/impl.h
gunes-arm: In another patch, can you add batch test cases for Fp16?

tests/validation/reference/ConvolutionLayer.cpp
```diff
@@ -107,31 +107,67 @@ SimpleTensor<T> convolution_layer_nchw(const SimpleTensor<T> &src, const SimpleT
...
-                                                 const PadStrideInfo &info, const Size2D &dilation, unsigned int num_groups, QuantizationInfo out_quant_info);
-template SimpleTensor<int8_t> convolution_layer(const SimpleTensor<int8_t> &src, const SimpleTensor<int8_t> &weights, const SimpleTensor<int32_t> &bias, const TensorShape &output_shape,
-                                                const PadStrideInfo &info, const Size2D &dilation, unsigned int num_groups, QuantizationInfo out_quant_info);
+// Dequantize i8+i8->F32
+template SimpleTensor<float> convolution_layer<int8_t,int8_t,float,float>(const SimpleTensor<int8_t> &src,
+                                                                              const SimpleTensor<int8_t> &weights, const SimpleTensor<float> &bias,
+                                                                              const TensorShape &output_shape,const PadStrideInfo &info, const Size2D &dilation,
+                                                                              unsigned int num_groups, QuantizationInfo out_quant_info);
+// Dequantize u8+u8->F32
```
gunes-arm, line 135: i8

morgolock: This is the template instatiation for u8, required to avoid compiler errors but not used.

PR #1185, opened 2025-09-23
Do only one iteration of newton-raphson refinement for FP16 inv

mcfi:
Do one iteration of newton-raphson refinement for FP16 inv
FP16 only has 10 explicitly stored bits of mantissa, so one
iteration of refinement should be precise enough. Numerical
results showed both one iteration and two iterations of
newton-raphson refinement had worst ULP being 1.

### DIFF
--- a/src/core/NEON/NEMath.inl
+++ b/src/core/NEON/NEMath.inl
@@ -634,14 +634,12 @@ inline float16x4_t vinv_f16(float16x4_t x)
 {
     float16x4_t recip = vrecpe_f16(x);
     recip             = vmul_f16(vrecps_f16(x, recip), recip);
-    recip             = vmul_f16(vrecps_f16(x, recip), recip);
     return recip;
 }
 
 inline float16x8_t vinvq_f16(float16x8_t x)
 {
     float16x8_t recip = vrecpeq_f16(x);
-    recip             = vmulq_f16(vrecpsq_f16(x, recip), recip);
     recip             = vmulq_f16(vrecpsq_f16(x, recip), recip);
     return recip;
 }
--- a/src/core/NEON/SVEMath.inl
+++ b/src/core/NEON/SVEMath.inl
@@ -82,7 +82,6 @@ inline svfloat16_t svinv_f16_z(svbool_t pg, svfloat16_t x)
 {
     auto recip = svrecpe_f16(x);
     recip      = svmul_f16_z(pg, svrecps_f16(x, recip), recip);
-    recip      = svmul_f16_z(pg, svrecps_f16(x, recip), recip);
     return recip;
 }
 


src/core/NEON/NEMath.inl
```diff
@@ -618,31 +618,27 @@ inline float16x4_t vinvsqrt_f16(float16x4_t x)
...
     sqrt_reciprocal = vmulq_f16(vrsqrtsq_f16(vmulq_f16(x, sqrt_reciprocal), sqrt_reciprocal), sqrt_reciprocal);
-    sqrt_reciprocal = vmulq_f16(vrsqrtsq_f16(vmulq_f16(x, sqrt_reciprocal), sqrt_reciprocal), sqrt_reciprocal);
     return sqrt_reciprocal;
 }
 
 inline float16x4_t vinv_f16(float16x4_t x)
 {
     float16x4_t recip = vrecpe_f16(x);
     recip             = vmul_f16(vrecps_f16(x, recip), recip);
-    recip             = vmul_f16(vrecps_f16(x, recip), recip);
```
gunes-arm, line 637: When/if accepted/verified, we should also change SVEMath.inl for it to be consistent across technologies.

mcfi:
Thanks for getting back, @gunes-arm, appreciated.

For inv, our results were the same. Yes, 2 iterations could get us slightly more precise results, but the worst case stays the same as 1 iteration. The question here is whether the extra iteration is better or worse. My take is, we should aim for the least iterations that get us worst ULP == 1, because the whole point of fast reciprocal is to to be fast with reasonably good precision. Otherwise, why not do 2+ iterations. :)

For invqrt, our results were different because I used the following for the reference values, which involves two roundings so slightly less precise than what you did by computing the double result and then rounding back to FP16. As your results showed, only 15 out  of 31743 cases had ULP == 2 while the remaining have ULP<=1. I would argue that the extra iteration for such a small precision gain is unnecessary but I wouldn't insist.

```
inline float16x4_t vinvsqrt_f16_div(float16x4_t x)
{
    float16x4_t recip = vdiv_f16(vdup_n_f16(1.0f), vsqrt_f16(x));
    return recip;
}
```
Does it sound good to you if I revert the invsqrt changes while keep the inv changes?

gunes-arm:
Hi @mcfi, my pleasure. I think we can certainly accept the contribution for the Inverse operation. You have a valid point there.  Unfortunately, the library does not set solid rules in numerical precision, but if we would have set such rules, I think worst-case ULP would have been that rule.

mcfi:
Sounds great, please see my latest iteration, thanks.

gunes-arm:
Hi Ben, the CI failed with commit type check. It should follow https://www.conventionalcommits.org/en/v1.0.0/ style (https://github.com/ARM-software/ComputeLibrary/blob/main/docs/contributor_guide/contribution_guidelines.dox#L340).

I believe `perf` is suitable. See our past commit messages for examples.

mcfi:
Oh, I see, please see my latest commit message. I installed the pre-commit hook and it seems like all the pre-commit checks have passed. Not sure why the jenkins' precommit job still failed. I can't see any details since I don't have access to the it...

gunes-arm:
Thanks. We trigger manually for external contributions. I've triggered it now. Let' see.

gunes-arm (approved):

mcfi:
Thanks, Gunes!

PR #1186, opened 2025-09-28
Sparse tensor data structures

aestriplex:
This PR implements a first draft of the data structures required to implement sparse tensors support in ACL. 
Here are three points worth discussing:
1. In the COOTensor and CSRTensor classes, I implemented two different strategies for saving indexes. Specifically, in COO tensors, indexes are saved as `std::vector` (i.e., not in the same buffer provided by the tensor allocator, which is used by values instead). In CSR tensors, on the other hand, I stored them in the allocator buffer and saved only the corresponding (integer) offsets as class members. The method used in CSR is more efficient when working with the tensor (e.g., in a kernel/operator), so I thought I would extend it to the COO tensor as well. Are there any problems if this approach is too “low level”?
2. The test class should be converted into a Fixture.
3. As Pablo said in his comment on my design document in the mailing list, we need to make sure that the validation of current operators fails when using sparse tensors. A good idea, in my opinion, would be to create a “validate” method in `ICpuOperator` - and other kind of base operators -, so that it can be called within the validate of classes that implement this interface and do _not_ (yet) support sparse tensors (e.g., CpuGemmConv2d). This method, i.e., not performing the direct check `!src->info()->is_sparse()`, allows these checks to be extended in the future without having to modify all operators.


arm_compute/runtime/COOTensor.h
```diff
@@ -0,0 +1,86 @@
...
+    COOTensor(const ITensor *tensor, size_t sparse_dim);
+    /** Convert a dense tensor to a *fully* sparse COOTensor.
+     *
+     *  @param[in] tensor
+     *
+     *  @note sparse_dim = tensor->info->num_dimensions()
+     */
+    COOTensor(const ITensor *tensor);
+
+    std::vector<Coordinates> _indices;
```
aestriplex: See point 1 in the PR description

arm_compute/runtime/CSRTensor.h
```diff
@@ -0,0 +1,92 @@
...
+     */
+    CSRTensor(const ITensor *tensor, size_t sparse_dim);
+    /** Convert a dense tensor to a *fully* sparse tensor.
+     *
+     *  @note sparse_dim = tensor->info->num_dimensions().
+     *        If tensor->info->num_dimensions() > 2 an error is raised.
+     */
+    CSRTensor(const ITensor *tensor);
+
+    size_t _crow_bytes; /**< Row index size in bytes */
```
aestriplex, line 84: See point 1 in the PR description

tests/validation/cpu/unit/SparseTensor.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+namespace test
+{
+namespace validation
+{
+TEST_SUITE(CPU)
+TEST_SUITE(UNIT)
+TEST_SUITE(SparseTensor)
+
+TEST_CASE(ConvertDenseTensorToCOOTensor, framework::DatasetMode::ALL)
```
aestriplex: This probably should become a Fixture test

arm_compute/core/IReducibleTensor.h
```diff
@@ -0,0 +1,76 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef ACL_ARM_COMPUTE_CORE_IREDUCIBLETENSOR_H
+#define ACL_ARM_COMPUTE_CORE_IREDUCIBLETENSOR_H
+
+#include "arm_compute/core/SparseTensor.h"
+#include "arm_compute/runtime/COOTensor.h"
+#include "arm_compute/runtime/CSRTensor.h"
```
morgolock: We should not include `runtime` headers in `core`

Can't we just forward declare these classes `COOTensor` and `CSRTensor`

src/runtime/COOTensor.cpp
```diff
@@ -0,0 +1,264 @@
...
+        for (size_t j = 0; j < coord.num_dimensions(); ++j)
+        {
+            os << coord[j];
+            if (j < coord.num_dimensions() - 1) os << ", ";
+        }
+        os << "]  values: ";
+        print_values(os, data, i, dense_volume(sparse_dim()));
+    }
+}
+#endif // ARM_COMPUTE_ASSERTS_ENABLED
```
morgolock, line 353: We already have a print method in https://github.com/ARM-software/ComputeLibrary/blob/main/src/core/ITensor.cpp

We should move this in `void ITensor::print(std::ostream &s, IOFormatInfo io_fmt) const`

src/runtime/CSRTensor.cpp
```diff
@@ -0,0 +1,272 @@
...
+        os << *reinterpret_cast<const int32_t *>(_col_indices + (i * index_size));
+        if (i < _col_bytes / index_size - 1)
+        {
+            os << ", ";
+        }
+    }
+    os << "] values: ";
+    print_values(os, values, 0, nnz());
+}
+#endif // ARM_COMPUTE_ASSERTS_ENABLED
```
morgolock, line 265: Move this to existing `void ITensor::print(std::ostream &s, IOFormatInfo io_fmt) const`

tests/validation/cpu/unit/SparseTensor.cpp
morgolock: Yes, we'll need to rework the test using fixtures and helpers in ACL

I'd say lets first start with simple unit tests like:
- https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/UNIT/TensorInfo.cpp
- https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/NEON/UNIT/TensorAllocator.cpp

The simple tests should create the new sparse tensors and test the new methods.

Then more complex tests with operators will require fixtures:
- https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/NEON/ArgMinMax.cpp
- https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/fixtures/ArgMinMaxFixture.h

src/runtime/SparseTensorAllocator.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "arm_compute/core/Error.h"
+#include "arm_compute/core/TensorInfo.h"
+#include "arm_compute/runtime/MemoryGroup.h"
+#include "arm_compute/runtime/MemoryRegion.h"
+
+#include <cstddef>
+
+namespace arm_compute
+{
+SparseTensorAllocator::SparseTensorAllocator(IMemoryManageable *owner) : _owner(owner), _associated_memory_group(nullptr), _memory(), _values_bytes(0), _indices_bytes(0)
```
morgolock, line 36: I think we should use the existing `TensorAllocator` to allocate any buffers.

morgolock (changes requested):
Hi @aestriplex 

I had a quick look at this and I think it's a good starting point. Please have a look and let me know your thoughts.

arm_compute/core/IReducibleTensor.h
aestriplex: I forward declared CSRTensor and COOTensor and removed the includes

tests/validation/cpu/unit/SparseTensor.cpp
aestriplex: I moved the conversion from/to sparse tensors tests in `validation/unit`, instead of `validation/cpu/unit`, and I used ACL helpers as done [https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/UNIT/TensorInfo.cpp](https://github.com/ARM-software/ComputeLibrary/blob/main/tests/validation/UNIT/TensorInfo.cpp)

src/runtime/COOTensor.cpp
aestriplex: I could move both this print function and CSR's to ITensor. In particular, something like this

```cpp
void ITensor::print(std::ostream &s, IOFormatInfo io_fmt) const
{
    // initialize variables and set precision
       ...
    switch (this->info()->tensor_format())
    {
        case TensorFormat::Dense:
            print_dense(s, io_fmt);
            break;
        case TensorFormat::CSR:
            print_csr(s);
            break;
        case TensorFormat::COO:
            print_coo(s);
            break;
        default:
            ARM_COMPUTE_ERROR("Unsupported tensor format");
            break;
    }

    // Restore output stream flags
    s.copyfmt(stream_status);
}
```

where `print_coo` and `print_csr` are the two functions moved from the two final classes.

The main problem is that those functions use some internals of the two classes, for example _indices, _crow_bytes, _col_bytes. It seems to me that having those functions defined in the base class can be misleading. Wouldn't it be cleaner if we have those two functions in their proper classes? That way, I should add the parameter `IOFormatInfo io_fmt` to both, so we have exactly the same signature as in ITensor, i.e. they overlap perfectly the base class method.

src/runtime/SparseTensorAllocator.cpp
aestriplex: I thought a lot about this one. The main reason I created a new class is that the original one (i.e. `TensorAllocator`) allocates the memory buffer using the `TensorShape` contained in `TensorInfo`. In the case of sparse tensors, however, the amount of memory that we need to allocate is smaller than that obtained by doing `dim_1 x dim_2 x ... x dim_n`, because otherwise there would be no advantage at all in using sparse over dense tensors.
Since I didn't want to touch the allocation mechanism of `TensorAllocator` — on which the memory allocation of all other tensors is based — I thought it would be wiser to create a special class. 
It has the same internal mechanism (i.e. it  inherits from  `ITensorAllocator`, and it uses `Memory`, `MemoryGroup` and `IMemoryManageable`), but it exposes this init method
```cpp
void init(const TensorInfo &input, size_t values_bytes, size_t indices_bytes, size_t alignment = 0);
```

aestriplex:
@morgolock Hi Pablo, this week I am going to continue implementing Transpose operator for both CSR and COO tensor. Are my fixes fine so far?
In particular, do you prefer indices to be represented as in CSRTensor (i.e., with allocators) or as in COOTensor (with C++ classes, i.e. std::vector)? Personally, I think the solution with allocators is by far the best in view of the implementation of operators. What do you think about it?

morgolock:
Hi @aestriplex are you planning on continue working on this PR? If so I'd suggest rebasing. Otherwise I'll close it.

aestriplex:
Hi @morgolock Yes, I am willing to continue working on this, so I'm doing a rebase.

But I was waiting for a feedback for https://github.com/ARM-software/ComputeLibrary/pull/1186#discussion_r2408934953 and https://github.com/ARM-software/ComputeLibrary/pull/1186#discussion_r2408950345, to avoid a potential rollback on those fundamental implementations

morgolock:
> Hi @morgolock Yes, I am willing to continue working on this, so I'm doing a rebase.
> 
> But I was waiting for a feedback for [#1186 (comment)](https://github.com/ARM-software/ComputeLibrary/pull/1186#discussion_r2408934953) and [#1186 (comment)](https://github.com/ARM-software/ComputeLibrary/pull/1186#discussion_r2408950345), to avoid a potential rollback on those fundamental implementations

Thanks @aestriplex , I'll have another look and provide feedback.

aestriplex:
Hi @morgolock I implemented the fixed for the copilot code review and added a few more commits  that implement:

1. More test cases
2. NHWC support for both CooTensor and CSRTensor

This completes the feature from a data structure point of view. Should I mark this PR as "ready for review" (i.e. exit from the "draft" stage)?

aestriplex:
> Hi @morgolock I implemented the fixed for the copilot code review and added a few more commits that implement:
> 
> 1. More test cases
> 2. NHWC support for both CooTensor and CSRTensor
> 
> This completes the feature from a data structure point of view. Should I mark this PR as "ready for review" (i.e. exit from the "draft" stage)?

@morgolock In addition to this, should I start implementing a Kernel for some basic operations, e.g. transpose?

PR #1192, opened 2025-10-17
tests: flatten nested combine usage in NEON validation

morgolock:
Change-Id: I976e01b9be9ee91f7de09f050392576133d1e929

### DIFF
--- a/tests/validation/NEON/SoftmaxLayer.cpp
+++ b/tests/validation/NEON/SoftmaxLayer.cpp
@@ -329,10 +329,9 @@ FIXTURE_DATA_TEST_CASE(RunSmall2D, NESoftmaxLayerQuantizedFixture<uint8_t>, fram
     combine(
         datasets::SoftmaxLayerSmallShapes(),
         make("DataType", DataType::QASYMM8),
-        combine(
             make("QuantizationInfo", { QuantizationInfo(0.5f, -10) }),
             make("Beta", { 1.0f, 2.f })
-        ),
+        ,
         make("Axis", { 0, -1 })))
 {
     // Validate output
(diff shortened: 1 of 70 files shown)


gunes-arm (commented):
I think the commit type should be `test`, but I guess I have a better suggestion. Isn't this a `refactor`?

gunes-arm:
Can you also count the number of tests before & after this patch as a sanity check, and report?

morgolock:
> I think the commit type should be `test`, but I guess I have a better suggestion. Isn't this a `refactor`?

I changed the commit to

```
refactor: flatten nested combine usage in NEON validation
```

morgolock:
> Can you also count the number of tests before & after this patch as a sanity check, and report?

With this patch `--mode=precommit`

> Executed 78213 test(s) (78213 passed, 0 expected failures, 0 failed, 0 crashed, 0 disabled) in 247 second(s)

Without this patch

> Executed 78213 test(s) (78213 passed, 0 expected failures, 0 failed, 0 crashed, 0 disabled) in 258 second(s)

gunes-arm:
Thanks for checking the number of tests. Can you check for --mode=ALL? You only need to list the tests, not run, i.e.
```
./arm_compute_validation --mode=ALL --list-tests | wc -l
```

tests/validation/NEON/SoftmaxLayer.cpp
```diff
@@ -383,10 +379,9 @@ FIXTURE_DATA_TEST_CASE(RunSmall4D, NESoftmaxLayerQuantizedFixture<int8_t>, frame
     combine(
         datasets::Small4DShapes(),
         make("DataType", DataType::QASYMM8_SIGNED),
-        combine(
```
gunes-arm, line 386: These ones are clever. I wish it could also amend the tab.

morgolock:
> Thanks for checking the number of tests. Can you check for --mode=ALL? You only need to list the tests, not run, i.e.
> 
> ```
> ./arm_compute_validation --mode=ALL --list-tests | wc -l
> ```

Both cases I get

```
130|shiba:/data/local/tmp/pabtel01 $ ./arm_compute_validation  --mode=ALL --list-tests | wc -l                                                                                        
425731

```

tests/validation/NEON/SoftmaxLayer.cpp
morgolock: We could certainly fix this in a second patch.

PR #1205, opened 2025-11-03
refactor: Flatten nested concat usage in validation/CL and CPP.

morgolock:
Updated CL/CPP validation tests to use the local make alias instead of fully qualifying framework::dataset::make.

Change-Id: I363d51ad9bbf15b2e972dda0a533f2929f3a4395


morgolock:
Both cases
```
./arm_compute_validation  --mode=ALL --list-tests | wc -l
488899
```

PR #1209, opened 2025-11-06
test: run clang-format in the tests folder

morgolock:
Change-Id: Icce2d4afe641b51672a977879b67a801d0ee3077


.pre-commit-config.yaml
```diff
@@ -103,7 +103,6 @@ repos:
         files: .*\.(cpp|h|hpp|hh|cc|inl|c)$
         exclude: |
             (?x)^(
```
gunes-arm: Let's just completely remove exclude section

morgolock: Done

PR #1212, opened 2025-11-10
feat: Add Single-ISA support and enable SME in Multi-ISA CMake build.

wajahat-abbas:
Resolves: COMPMID-8547

Change-Id: I82d3e74b28284a4052a2042f47d2eff896cc9c38
Signed-off-by: Syed Wajahat Abbas Naqvi <syedwajahatabbas.naqvi@arm.com>


tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -1032,7 +1033,8 @@ FIXTURE_DATA_TEST_CASE(UC2_2_NEGEMMConvolutionLayer_FastMath,
                        combine(make("DataType", {DataType::F32}),
                                make("QueryWeightFormat", {arm_compute::WeightFormat::OHWIo8i4_bf16})))
 {
-    if (Scheduler::get().cpu_info().has_bf16() && (arm_gemm::utils::get_vector_length<float>() == 8))
+    if (Scheduler::get().cpu_info().has_bf16() && Scheduler::get().cpu_info().has_sve() &&
```
gunes-arm, line 1036: Why did we need to change here?

cmake/compilers/setup.cmake
```diff
@@ -41,26 +41,56 @@ set(
...
+    ARM_COMPUTE_ENABLE_SVE2
+    ARM_COMPUTE_ENABLE_SVEF32MM
+    ENABLE_SVE
+    ENABLE_SVE2
+    ENABLE_FP16_KERNELS
+    ARM_COMPUTE_ENABLE_FIXED_FORMAT_KERNELS
+  )
+
+else()
+  if(ACL_ARCH_ISA STREQUAL "arm64-v8a")
```
gunes-arm: I think we should cover other synonymus options in these if/else's, such as armv8a here.

In essence, we should cover everything mentioned in Sconstruct here:

```
EnumVariable("arch", "Target Architecture. The x86_32 and x86_64 targets can only be used with neon=0 and opencl=1.", "armv7a",
                  allowed_values=("armv7a", "armv7a-hf", "arm64-v8a", "arm64-v8.2-a", "arm64-v8.2-a-sve", "arm64-v8.2-a-sve2", "x86_32", "x86_64",
                                  "armv8a", "armv8.2-a", "armv8.2-a-sve", "armv8.6-a", "armv8.6-a-sve", "armv8.6-a-sve2", "armv8.6-a-sve2-sme2", "armv8r64", "
```

The goal is being able to call CMake from Scons and obtain the same behaviour.

tests/validation/NEON/ConvolutionLayer.cpp
wajahat-abbas: It was causing illegal instruction exception in Fixed Format Kernel tests if not guarded with `has_sve`

cmake/compilers/setup.cmake
wajahat-abbas: Sure, I can look into this however the design doc specifically mentioned three presets: `arm64-v8a, arm64-v8.2-a, arm64-v8.6-a.`

gunes-arm: Actually, let's rethink this. We can ignore 32-bit and armv7a ones as we're doing this for macOS in the scope of this epic. Same goes for armv8r64, x86 etc.

One simple thing to do is to get rid of the "64" bit in the arch names. So, CMake should take armv8-a, armv8.2-a etc. without the 64 in them. When calling from Scons (in the future), we can convert all of them to CMake compatible ones.

We should also be aware any of the misdesigns in the previous implementation (i.e. Scons) and address properly as we have the chance now. One of those misdesigns is to explicitly spell out each arch and feature combination in the supported archs, such as armv8.2-a-sve, armv8.6-a-sve, armv8.6-a-sve2-sme2. arch and feature are combined in the same argument and there are combinations that we can support but don't, and for each one of them we have to add a parameter option. I think we can and should do things differently when building the single-isa solution in CMake. So, I'm taking my initial suggestion back and revising it. How about we get the feature names above (we can start with sve, sve2 and sme2) as a comma separated list (or whatever you think is better), set the arch based on that and add the flags.

wajahat-abbas: Good suggestion. Does this mean that we do not need arch as input at all and instead deduce it based on feature names?

gunes-arm: Commenting here for completeness, we won't be decuding arch from the features. My last sentence in the above logn comment is a bit unfortunate :) I'm sorry for the confusion I caused. As agreed offline, we'll still specify the base arch. but take the feature list in some way (e.g. comma-separated) and build the -march compiler argument based on that. This will ensure scalability in the future.

cmake/compilers/setup.cmake
```diff
@@ -41,26 +41,98 @@ set(
...
   ENABLE_FP32_KERNELS
   ENABLE_QASYMM8_KERNELS
   ENABLE_QASYMM8_SIGNED_KERNELS
   ENABLE_QSYMM16_KERNELS
   ENABLE_INTEGER_KERNELS
   ENABLE_NHWC_KERNELS
   ENABLE_NCHW_KERNELS
+)
+
+if(ACL_MULTI_ISA)
```
morgolock, line 54: It would be nice to have a message printing all the features enabled in the multi_isa build. So that we don't have to look into the makefiles to see if a specific feature has been enabled

wajahat-abbas: Message is printed but it doesn't explicitly say what features are enabled: "-- Multi-ISA selected, requested features list will be ignored."

wajahat-abbas: I will add a message printing features. Should it be a high level one e.g. lists only sve, sve2 etc or do we need to print other macros as well e.g. FP16, BF16 etc?

morgolock: High level is fine, we just need someone who does not know the build system or much about ACL to know what cpu feat are enable in multi_isa.

CMakePresets.json
```diff
@@ -22,6 +22,20 @@
                 "ARM_COMPUTE_BUILD_TESTING": "ON"
             }
         },
+        {
```
gunes-arm: I don't think we need a preset for single-isa and features. There will always be a question of what to add here. Best to not not touch at all.

CMakeLists.txt
```diff
@@ -49,11 +49,24 @@ option(ARM_COMPUTE_ENABLE_CODE_COVERAGE "Enable code coverage." OFF)
...
+option(ACL_BUILD_SVE2 "Enable SVE2 support" OFF)
+option(ACL_BUILD_SME2 "Enable SME2 support" OFF)
+
+# * Arch feature probing for single-ISA
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag("-march=armv8-a"                              ACL_HAS_MARCH_V8A)
+check_cxx_compiler_flag("-march=armv8.2-a+fp16+dotprod"               ACL_HAS_MARCH_V82_F16_DOT)
+check_cxx_compiler_flag("-march=armv8.6-a+sve+sve2+fp16+dotprod+i8mm" ACL_HAS_MARCH_V86_ALL)
+check_cxx_compiler_flag("-march=armv8.2-a"                            ACL_HAS_MARCH_V82_BASE)
+check_cxx_compiler_flag("-march=armv8.6-a"                            ACL_HAS_MARCH_V86_BASE)
```
gunes-arm, line 69: We're not using these vars, right?

CMakeLists.txt
```diff
@@ -72,6 +85,57 @@ include(${CMAKE_CURRENT_LIST_DIR}/cmake/configurations.cmake)
...
+if(ACL_ARCH_ISA STREQUAL "armv8-a" AND (ACL_MULTI_ISA OR ACL_BUILD_SVE OR ACL_BUILD_SVE2 OR ACL_BUILD_SME2))
+  message(FATAL_ERROR
+    "Invalid configuration: '${ACL_ARCH_ISA}' does not support requested features"
+  )
+endif()
+
+# * Print build configuration
+if(ACL_MULTI_ISA)
+  message(STATUS "Multi-ISA build selected.")
+  message(STATUS "Building ACL libraries: core, core_fp16, sve, sve2.")
```
gunes-arm: also sme2?

CMakeLists.txt
wajahat-abbas: This is just to show message if compiler supports those -march or not.

CMakeLists.txt
wajahat-abbas: sme2 sources are part of sve and sve2 libs since sme depends on it. Please see: https://github.com/ARM-software/ComputeLibrary/blob/main/src/CMakeLists.txt

gunes-arm: Okay, you were referring to the CMake variables/libraries. I thought this message was listing the features. To be more clear, I suggest
message(STATUS "Multi-ISA build selected. All detectable features are enabled.")
message(STATUS "Building arm_compute_(core, core_fp16, sve, sve2).") --> or a better reading one :)

CMakeLists.txt
gunes-arm: Resolving this. I didn't know it was printing anything. Then, it's useful.

CMakeLists.txt
wajahat-abbas: The message will look like this, here are some examples:
```
-- Multi-ISA build selected.
-- Building ACL libraries: core, core_fp16, sve, sve2

-- Single-ISA build selected: armv8.2-a
-- Requested features: sme2
-- Building ACL libraries: core, core_fp16, sve, sve2.

-- Single-ISA build selected: armv8-a
-- Building ACL libraries: core.

-- Single-ISA build selected: armv8.6-a
-- Requested features: sve
-- Building ACL libraries: core, core_fp16, sve.

-- Single-ISA build selected: armv8.6-a
-- Requested features: sve,sve2,sme2
-- Building ACL libraries: core, core_fp16, sve, sve2.
```

gunes-arm: Thanks, I still think people who are not familiar with how our CMake and build system works can wonder if they enabled other HW features, too.

I was confused by "core, core_fp16, sve, sve2", so can others.

tests/validation/NEON/ConvolutionLayer.cpp
gunes-arm: Thanks, looks like we're not testing fixed format kernels in a non-SVE device. Luckily, the issue is only on test code.

wajahat-abbas: Do you have any changes to suggest here or this one is okay?

gunes-arm: No, I resolved the thread.

gunes-arm: It's out of scope for this work.

CMakeLists.txt
morgolock: If sme1 and sme2 are enabled we should inform it with a message

wajahat-abbas: Currently Multi-ISA configuration does not enables the SME flags and I have sticked with that in this PR. Should those be enabled for Multi-ISA?

morgolock: I think so, that's the point of muli_isa, to build all the kernels and select at runtime

wajahat-abbas: I understand however its not enabled in existing codebase so perhaps it was intentionally left out?

morgolock: I think there are more use cases now for us to enable this, firstly there is silicon with sme and secondly there is GBML which could benefit from it.

It could be enabled in another patch /PR.

gunes-arm: If you're talking about -march flags, they are left out intentioanally because the compliation doesn't need them as the instructions are in opcode. If you're referring to the message we give here, that's something else though.

morgolock: Surely we need to enable `ARM_COMPUTE_ENABLE_SME2` to compile in the SME2 kernels in multi_isa?

wajahat-abbas: @gunes-arm I am referring to `ARM_COMPUTE_ENABLE_SME` `ARM_COMPUTE_ENABLE_SME2` `ENABLE_SME`. These are not set in existing codebase which is Multi-ISA.

morgolock: > @gunes-arm I am referring to `ARM_COMPUTE_ENABLE_SME` `ARM_COMPUTE_ENABLE_SME2` `ENABLE_SME`. These are not set in existing codebase which is Multi-ISA.

Then it's disabled currently, we should enable this for multi_isa.

gunes-arm: Good catch then. All the flags should already be enabled in the current multi-isa builds. At least, in SCons, I know they're enabled. @wajahat-abbas are you referring to CMake multi-isa or SCons multi-isa for macOS?

wajahat-abbas: @gunes-arm CMake Multi-ISA.

gunes-arm: Okay, that's wrong and should be fixed. We should enable all the flags. But, it makes more sense to do it in another patch with a dedicated commit title

PR #1215, opened 2025-11-17
feat: Provide libarm_compute.a in Linux(R) Cpu only multi-isa build v…

gunes-arm:
…ia scons

Duplicate libarm_compute-static.a with name libarm_compute.a. libarm_compute-static.a will be removed in future releases, and libarm_compute.a will be used instead in all applicable releases.

This new binary will be accessible in build time only. It will be available in pre-built release binaries after it's been deprecated and rolled out to other builds with a major version upgrade.

Partially Resolves: COMPMID-8695

Change-Id: I59c3e809fc282d6dd16071184203e5be1d4af4d6

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -744,6 +744,13 @@ if (env['multi_isa']):
     lib_static_objs, lib_shared_objs = build_multiisa_lib_objects()
     # STATIC library build.
     arm_compute_a = build_library('arm_compute-static', arm_compute_env, lib_static_objs, static=True)
+
+    if(env['os'] == 'linux' and not env['opencl']):
+        # -static suffix in static libraries is deprecated, and will be removed in COMPMID-8696.
+        # This additional static binary is provided temporarily for multi-isa builds upon user request,
+        # and after the deprecation notice, the version with -static suffix will be removed in a
+        # future major release.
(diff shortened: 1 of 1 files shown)


SConscript
```diff
@@ -744,6 +744,13 @@ if (env['multi_isa']):
     lib_static_objs, lib_shared_objs = build_multiisa_lib_objects()
     # STATIC library build.
     arm_compute_a = build_library('arm_compute-static', arm_compute_env, lib_static_objs, static=True)
+
+    if(env['os'] == 'linux' and not env['opencl']):
+        # -static suffix in static libraries is deprecated, and will be removed in COMPMID-8696.
+        # This additional static binary is provided temporarily for multi-isa builds upon user request,
+        # and after the deprecation notice, the version with -static suffix will be removed in a
+        # future major release.
+        arm_compute_a_wo_suffix = build_library('arm_compute', arm_compute_env, lib_static_objs, static=True)
```
morgolock, line 753: so we will be generate `libarm_compute.a` from now on instead of `libarm_compute-static.a`?

gunes-arm: For now, this will generate both. A patch after this will put a deprecation notice and we'll rename everything after some time.

PR #1218, opened 2025-11-18
fix: Exclude the FP16 MMUL Reshaped RHS kernel for MMUL extension version less than 1.1.0

ArmDude:
Partially Resolves: [COMPMID-8596]

Change-Id: I98fb24ac420f978c8456596493fc4013090676cb

### DIFF
--- a/src/core/CL/CLHelpers.cpp
+++ b/src/core/CL/CLHelpers.cpp
@@ -289,6 +289,40 @@ bool device_supports_extension(const cl::Device &device, const char *extension_n
     return (pos != std::string::npos);
 }
 
+bool device_supports_extension_version(const cl::Device &device,
+                                       const char       *extension_name,
+                                       unsigned int      major_required,
+                                       unsigned int      minor_required,
+                                       unsigned int      patch_required = 0)
+{
+    const auto extensions = device.getInfo<CL_DEVICE_EXTENSIONS_WITH_VERSION>();
+
+    for (const auto &ext : extensions)
+    {
+        if (std::strcmp(extension_name, ext.name) == 0)
+        {
+            unsigned int major = CL_VERSION_MAJOR(ext.version);
+            unsigned int minor = CL_VERSION_MINOR(ext.version);
+            unsigned int patch = CL_VERSION_PATCH(ext.version);
+
+            // The device supports the extension if its version >= required
+            if (major > major_required)
+                return true;
+            if (major < major_required)
+                return false;
+
+            if (minor > minor_required)
+                return true;
+            if (minor < minor_required)
+                return false;
+
+            return patch >= patch_required;
+        }
+    }
+
+    return false;
+}
+
 bool cl_winograd_convolution_layer_supported(const Size2D &output_tile,
                                              const Size2D &kernel_size,
                                              DataLayout    data_layout)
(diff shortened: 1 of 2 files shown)


src/core/CL/CLHelpers.cpp
```diff
@@ -289,6 +289,26 @@ bool device_supports_extension(const cl::Device &device, const char *extension_n
     return (pos != std::string::npos);
 }
 
+bool device_supports_extension_version(const cl::Device &device, const char *extension_name, unsigned major_required, unsigned minor_required, unsigned patch_required = 0)
+{
+    auto extensions = device.getInfo<CL_DEVICE_EXTENSIONS_WITH_VERSION>();
```
gunes-arm: const

src/core/CL/CLHelpers.cpp
```diff
@@ -508,11 +528,8 @@ bool arm_matrix_multiply_fp16_supported(const cl::Device &device)
         return false;
     }
 
-    cl_ulong   caps = 0;
```
gunes-arm, line 511: We should keep hw capability checks in addition to the extension version

src/core/CL/CLHelpers.cpp
ArmDude: Done!

src/core/CL/CLHelpers.cpp
ArmDude: How's this?

src/core/CL/CLHelpers.cpp
```diff
@@ -512,7 +545,9 @@ bool arm_matrix_multiply_fp16_supported(const cl::Device &device)
     const auto err  = clGetDeviceInfo( //
         device(), CL_DEVICE_MATRIX_MULTIPLY_CAPABILITIES_ARM, sizeof(caps), &caps, nullptr);
 
-    const auto supported = err == CL_SUCCESS && (caps & CL_DEVICE_MATRIX_MULTIPLY_FP16_WITH_FP16_ACCUMULATORS_ARM) != 0;
+    const auto supported = err == device_supports_extension_version(device, "cl_arm_matrix_multiply", 1, 1, 0) &&
```
gunes-arm: I think this should be (err == CL_SUCCESS) && device_supports...() && (caps & CL_DEVICE_MATRIX_MULTIPLY_FP16_WITH_FP16_ACCUMULATORS_ARM) != 0

src/core/CL/CLHelpers.cpp
```diff
@@ -289,6 +289,39 @@ bool device_supports_extension(const cl::Device &device, const char *extension_n
...
+                                       unsigned          minor_required,
+                                       unsigned          patch_required = 0)
+{
+    const auto extensions = device.getInfo<CL_DEVICE_EXTENSIONS_WITH_VERSION>();
+
+    for (const auto &ext : extensions)
+    {
+        if (std::strcmp(extension_name, ext.name) == 0)
+        {
+            unsigned major = CL_VERSION_MAJOR(ext.version);
```
gunes-arm: `unsigned` per se is not a conventional use. Same goes for function arguments. What is exactly being returned from ext.version? cl_uint?

ArmDude: It's a cl_uint which guarantees portability across different host platforms, it's a 32 bit unsigned type. Shall I go for unsigned int instead?

gunes-arm: Let's use `unsigned int`

src/core/CL/CLHelpers.cpp
ArmDude: How's this?

src/core/CL/CLHelpers.cpp
ArmDude: How's this?

PR #1231, opened 2025-12-12
test: Adjust chosen scales in tests for Int8 x Int8 -> f32 Convolution

gunes-arm:
Int8 x Int8 -> f32 convolution is picking random scales as if the output is Int8/UInt8, too. In fully quantized convolutions, given input & weight quantizations, we pick the dst quantization info in such a way that the results don't saturate in the 8-bit integer representation.

In f32 outputs, the bias is f32, too, and we don't have a dst quant. info to adjust. In this case, randomly chosen input & weight quant. infos might cause large multiplication outputs. But, the random bias is not chosen according to this, and this causes rounding errors.

The situation usually happens when one of the output values w/o added bias is close to 0. Because the large multiplication output is cancelled by the offset contribution, which makes the output 0, and the result should be equal to bias. However, the bias is added in the assembly kernel before the offset contribution is added to the multiplication output. If the multiplication output is much larger than the bias, it results in big rounding errors int the final sum. When the offset contribution cancels the multiplication output, the rounded bias becomes equal to the result. However, the rounding error sometimes exceeds the tolerance in such cases. This is not healthy as increasing tolerance could result in ignoring the bias and future errors related to that all together.

This commit adjusts the input and weight quantization info to prevent such cases.

Resolves: COMPMID-8660

Change-Id: I554b6c9faabfdc1d3b755487dbe253f9658d03e2

### DIFF
--- a/tests/validation/fixtures/ConvolutionLayerFixture.h
+++ b/tests/validation/fixtures/ConvolutionLayerFixture.h
@@ -124,29 +124,59 @@ public:
                             QuantizationInfo &weights_q_info,
                             DataType          data_type)
     {
-        const int32_t t_max = static_cast<int32_t>(std::numeric_limits<T>::max());
-        const int32_t t_min = static_cast<int32_t>(std::numeric_limits<T>::min());
+        std::mt19937 generator(library->seed() + _hash);
 
-        std::mt19937                           generator(library->seed() + _hash);
-        std::uniform_real_distribution<float>  distribution_float(-5.0f, 3.0f);
+        const int32_t                          t_max = static_cast<int32_t>(std::numeric_limits<T>::max());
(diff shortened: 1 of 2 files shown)


tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -623,7 +654,7 @@ class ConvolutionValidationGenericFixture : public framework::Fixture
         auto conv = reference::convolution_layer<T, TW, TBias, TO>(src, weights, bias, output_shape, info, dilation,
                                                                    num_groups, _dst_q_info);
         auto res  = (act_info.enabled()) ? reference::activation_layer<TO>(conv, act_info) : conv;
-
+        // std::cout << res[0] << std::endl;
```
morgolock: remove

morgolock (approved):

PR #1278, opened 2026-04-10
perf: Enable assembly kernel for QASYMM8 MAX pooling.

morgolock:
Change-Id: I96cdf94dc29348a01b8f47aba1b6afd6d82fbcb9

### DIFF
--- a/src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
+++ b/src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2025 Arm Limited.
+ * Copyright (c) 2021-2026 Arm Limited.
  *
  * SPDX-License-Identifier: MIT
  *
@@ -145,9 +145,10 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
             if (src->data_type() == DataType::QASYMM8)
             {
                 const bool has_padding = info.pad_stride_info.has_padding();
-                ARM_COMPUTE_RETURN_ERROR_ON_MSG(
-                    !info.exclude_padding && has_padding,
-                    "Assembly kernels do not support padding for QASYMM8 with same src/dst quantization info");
+                ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX && !info.exclude_padding &&
+                                                    has_padding,
+                                                "Assembly kernels only support padded MAX pooling for QASYMM8 with "
+                                                "same src/dst quantization info");
             }
         }
     }
@@ -159,9 +160,9 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
         {
             // If dst is not configured, the quantization info are the same
             const bool has_padding = info.pad_stride_info.has_padding();
-            ARM_COMPUTE_RETURN_ERROR_ON_MSG(
-                !info.exclude_padding && has_padding,
-                "Assembly kernels do not support padding for QASYMM8 with same src/dst quantization info");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX && !info.exclude_padding && has_padding,
+                                            "Assembly kernels only support padded MAX pooling for QASYMM8 with "
+                                            "same src/dst quantization info");
         }
     }
     return Status{};
--- a/tests/validation/runtime/experimental/operators/CpuPool2d.cpp
+++ b/tests/validation/runtime/experimental/operators/CpuPool2d.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2021, 2023-2025 Arm Limited.
+ * Copyright (c) 2017-2021, 2023-2026 Arm Limited.
  *
  * SPDX-License-Identifier: MIT
  *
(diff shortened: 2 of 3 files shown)


src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
```diff
@@ -146,8 +146,8 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
             {
                 const bool has_padding = info.pad_stride_info.has_padding();
                 ARM_COMPUTE_RETURN_ERROR_ON_MSG(
-                    !info.exclude_padding && has_padding,
-                    "Assembly kernels do not support padding for QASYMM8 with same src/dst quantization info");
+                    info.pool_type == PoolingType::AVG && !info.exclude_padding && has_padding,
```
gunes-arm: I think checking for != MAX is easier. This sentence alone implies L2 pooling gets along well with paddings. So, checking for types that we know they're supported conveys the message more clearly.

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
gunes-arm, line 145: I think the speciality around `QASYMM8` is that the minimum value is `0` and the initial value is set to this. So, `!exclude_padding` is not supported when the type is `QASYMM8_SIGNED` as the minimum value must be `-128`. I wonder if it's an easy piece of code that we can change to support both types. Given that it's only `QASYMM8` for now, we should state it in the commit message as is, not as `Int8`.

gunes-arm, line 141: This part is wildly weird because it looks like we don't check for all the padding related issues if the quantization informations are different. What is going on here?

tests/validation/runtime/experimental/operators/CpuPool2d.cpp
gunes-arm, line 1: I don't think this is the right place to add the test. `tests/validation/NEON/PoolingLayer.cpp` is the correct place. I'd not call it as a Smoke test as well. It's a proper test imo.

gunes-arm (changes requested):
Regarding the commit message, I think we should add the tickets and the GitHub issue number.

Question to you: Is this a fix or something else? Looks like a performance improvement to me.

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: Updated the commit msg to mention just `QASYMM8` for now.

I'll look into adding support for `QASYMM8_SIGNED` in a different patch to keep this change small.

tests/validation/runtime/experimental/operators/CpuPool2d.cpp
morgolock: We already had Smoke tests in this file, that's why I used the name.

I will move the new test to `tests/validation/NEON/PoolingLayer.cpp`

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: Good point, I'm not sure, I can investigate this too in a different patch. I don't want to increase the complexity of this small patch fixing the perf regression.

gunes-arm: Can you also update the PR title and description? I also have another comment regarding the commit category, which is around the top of the main conversation panel.

gunes-arm: What we need to do to at least add the same if/else here and add tests. I'm ok to do it in another patch if we handle it right after that. We probably also need a COMPMID Bug ticket for this.

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
gunes-arm: It marks this one as outdated, but the comment still stands.

tests/validation/runtime/experimental/operators/CpuPool2d.cpp
gunes-arm: Yes, they were called as smoke tests because what they were testing is just the plumbing from Op API to internal Op API. The complete functionality testing is from the official API. 

If the public experimental Op API is removed in the future, these tests would disappear.

gunes-arm: Resolving this one. Thanks for adding them here.

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: I created the tickets below for the additional investigation work:
- MLCE-1821
- MLCE-1822

I'll update the commit msg and pr title to `perf: Enable assembly kernel for QASYMM8 MAX pooling`

morgolock: I created MLCE-1821 to investigate this. Do we also want a COMPMID bug?

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: Addressed in next patchset

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
gunes-arm: I guess there is no need. There's been a mistake here but two mistakes cancelled each other and there is no actually driveable bug here. Resolving.

gunes-arm: PR title is still the same, isn't it?

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
```diff
@@ -159,9 +160,9 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
         {
             // If dst is not configured, the quantization info are the same
             const bool has_padding = info.pad_stride_info.has_padding();
-            ARM_COMPUTE_RETURN_ERROR_ON_MSG(
-                !info.exclude_padding && has_padding,
-                "Assembly kernels do not support padding for QASYMM8 with same src/dst quantization info");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX && !info.exclude_padding && has_padding,
+                                            "Assembly kernels only support padded AVG pooling for QASYMM8 with "
```
gunes-arm: The comment should probably slightly change to reflect the `!= MAX` check. Such as

> Assembly kernels only support MAX pooling if there is padding and exclude_padding is false.

morgolock: Missed it, will update in next patchset

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: Thanks, done now.

gunes-arm (approved):

PR #1279, opened 2026-04-15
feat: Enable BF16 I/O for CpuFullyConnected in the experimental Operator API

### LINKED ISSUE #180447: Poor BF16 torch.compile + Freezing perf on AArch64 CPUs
### 🐛 Describe the bug

When you compile and freeze a BF16 model (for weight prepacking) on AArch64, the matmul gets dispatched to oneDNN reference kernels rather than ACL/jit optimized oneDNN kernels. As a result you run 100x slower than eager mode.

This is reproducible with any torch version.
Note: for fp32, we already dispatch to optimized kernels.

```
import torch 

torch._inductor.config.freezing = True

# This dispatches to optimized oneDNN kernels
#dtype = torch.float32

# This dispatches to reference oneDNN kernels -> more than 100x slower than eager 
dtype = torch.bfloat16

def main():
    with torch.no_grad():
        x = torch.rand(size=(1024, 1024), dtype=dtype)
        linear = torch.nn.Linear(1024, 1024).to(dtype).eval()
        # compile and warmup
        linear = torch.compile(linear)
        linear(x)
        print("should have no reorders after this", flush=True)
        for _ in range(10):
            linear(x)
            print("=="*20, flush=True)

if __name__ == "__main__":
    main()
```

cc @jerryzh168 @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @aditew01 @gujinghui @PenghuiCheng @jianyuh @min-jean-cho @yanbing-j @Guobing-Chen @Xia-Weiwen @snadampal @milpuz01 @nikhil-arm @robert-hardwick @nWEIdia @chauhang @penguinwu @voznesenskym @EikanWang @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @aakhundov @coconutruben @jataylo

fadara01:
feat: Enable BF16 CpuFullyConnected

Support is already there, so just removed defensive checks
Partially fixes: https://github.com/pytorch/pytorch/issues/180447

### DIFF
--- a/src/runtime/experimental/operators/CpuFullyConnected.cpp
+++ b/src/runtime/experimental/operators/CpuFullyConnected.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Arm Limited.
+ * Copyright (c) 2025-2026 Arm Limited.
  *
  * SPDX-License-Identifier: MIT
  *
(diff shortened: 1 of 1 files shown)


gunes-arm:
I suggest

```
feat: Enable BF16 I/O for CpuFullyConnected in the experimental Operator API
```

as the commit title. Also in the description, it sounds like "these" should be "there".

gunes-arm (dismissed):

gunes-arm (approved):

PR #1282, opened 2026-04-20
test: Relax pre-commit sign-off checkers for external contributions

gunes-arm:
Change-Id: I53a232b6f9c0de160ffe2ad849a11fd6158dc079

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -100,9 +100,12 @@ repos:
         stages: [pre-push]
       - id: check-signoff
         name: Check Signed-off-by
-        entry: bash -c 'grep -qi "Signed-off-by:.*@arm.com" "$0" || { echo "Missing Signed-off-by with @arm.com"; exit 1; }'
+        entry: >
+          bash -c 'git log -1 --pretty=%B |
+          grep -qiE "^Signed-off-by: .+ <.+@.+>$"'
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -100,9 +100,12 @@ repos:
         stages: [pre-push]
       - id: check-signoff
         name: Check Signed-off-by
-        entry: bash -c 'grep -qi "Signed-off-by:.*@arm.com" "$0" || { echo "Missing Signed-off-by with @arm.com"; exit 1; }'
+        entry: >
+          bash -c 'git log -1 --pretty=%B |
+          grep -qiE "Signed-off-by:\s+.*<[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}>$"'
```
andflo-Arm: Currently this would fail on say foo@mail.example.com. Would it suffice to check that there is a line starting with "Signed-off-by:" rather than trying to parse valid emails?

andflo-Arm (changes requested):

.pre-commit-config.yaml
gunes-arm: Removed the strict check and just added basic ones.

andflo-Arm (approved):

PR #1285, opened 2026-04-27
fix: Enable asm requantized max pooling with differing qinfo

morgolock:
Restrict differing src/dst quantization-info support in CpuPool2dAssemblyWrapperKernel to MAX pooling, while keeping AVG pooling on the generic fallback path.

Fix the quantized multiplier validation check and wire the requantization shifts correctly for the asm pooling path so requantized MAX pooling validates and executes correctly.

Add NEON validate coverage for padded NHWC QASYMM8 MAX with differing quantization info.

Resolves MLCE-1821

Change-Id: I33c99d0d4ea1bf57ed28d0750422403e60e2a276

### DIFF
--- a/src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
+++ b/src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
@@ -134,21 +134,28 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
 
         if (src_qinfo != dst_qinfo)
         {
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX,
+                                            "Assembly kernels only support differing src/dst quantization info for "
+                                            "MAX pooling");
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(src->data_type() == DataType::QASYMM8_SIGNED &&
+                                                (src_qinfo.offset != 0 || dst_qinfo.offset != 0),
+                                            "Assembly kernels only support differing src/dst quantization info for "
+                                            "QASYMM8_SIGNED when both offsets are zero");
             const float multiplier = src_qinfo.scale / dst_qinfo.scale;
             int32_t     dst_multiplier{};
             int32_t     dst_shift{};
             ARM_COMPUTE_RETURN_ERROR_ON(
-                quantization::calculate_quantized_multiplier(multiplier, &dst_multiplier, &dst_shift));
+                !quantization::calculate_quantized_multiplier(multiplier, &dst_multiplier, &dst_shift));
         }
         else
         {
-            if (src->data_type() == DataType::QASYMM8)
+            if (src->data_type() == DataType::QASYMM8 || src->data_type() == DataType::QASYMM8_SIGNED)
             {
                 const bool has_padding = info.pad_stride_info.has_padding();
                 ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX && !info.exclude_padding &&
                                                     has_padding,
-                                                "Assembly kernels only support padded MAX pooling for QASYMM8 with "
-                                                "same src/dst quantization info");
+                                                "Assembly kernels only support padded MAX pooling for QASYMM8 and "
+                                                "QASYMM8_SIGNED with same src/dst quantization info");
             }
         }
     }
@@ -156,13 +163,13 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
     {
         const TensorInfo out_info(dst_shape, 1, src->data_type()); // use src dtype as they're same.
         ARM_COMPUTE_RETURN_ERROR_ON_SIZE_UNSUPPORTED(&out_info);
-        if (src->data_type() == DataType::QASYMM8)
+        if (src->data_type() == DataType::QASYMM8 || src->data_type() == DataType::QASYMM8_SIGNED)
         {
             // If dst is not configured, the quantization info are the same
             const bool has_padding = info.pad_stride_info.has_padding();
             ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX && !info.exclude_padding && has_padding,
-                                            "Assembly kernels only support padded MAX pooling for QASYMM8 with "
-                                            "same src/dst quantization info");
+                                            "Assembly kernels only support padded MAX pooling for QASYMM8 and "
+                                            "QASYMM8_SIGNED with same src/dst quantization info");
         }
     }
     return Status{};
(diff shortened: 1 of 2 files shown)


src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
```diff
@@ -134,11 +134,14 @@ CpuPool2dAssemblyWrapperKernel::validate(const ITensorInfo *src, const ITensorIn
 
         if (src_qinfo != dst_qinfo)
         {
+            ARM_COMPUTE_RETURN_ERROR_ON_MSG(info.pool_type != PoolingType::MAX,
```
gunes-arm, line 137: Don't we need to check for type, too?

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
gunes-arm: We seem to accept qasymm8_signed, I think that wasn't the case.

tests/validation/NEON/PoolingLayer.cpp
```diff
@@ -157,13 +159,25 @@ DATA_TEST_CASE(Validate, framework::DatasetMode::ALL, zip(
...
-    make("Expected", { false, false, false, false, true, false, true, false, false, false, false})
+    make("Expected", { false, false, false, false, true, false, true, false, false, false, false, true})
     ),
     input_info, output_info, pool_info, expected)
 {
-    bool is_valid = bool(NEPoolingLayer::validate(&input_info.clone()->set_is_resizable(false), &output_info.clone()->set_is_resizable(false), pool_info));
+    TensorInfo input_info_configured  = input_info;
+    TensorInfo output_info_configured = output_info;
+
+    if(input_info_configured.tensor_shape() == TensorShape(3U, 15U, 11U, 1U))
```
gunes-arm: This looks a bit hacky, I would rather prefer a simple `TEST_CASE` for this.

tests/validation/NEON/PoolingLayer.cpp
```diff
@@ -157,13 +159,25 @@ DATA_TEST_CASE(Validate, framework::DatasetMode::ALL, zip(
...
+    {
+        input_info_configured.set_quantization_info(QuantizationInfo(0.25f, 11));
+        output_info_configured.set_quantization_info(QuantizationInfo(0.5f, 7));
+    }
+
+    bool is_valid = bool(NEPoolingLayer::validate(&input_info_configured.clone()->set_is_resizable(false),
+                                                  &output_info_configured.clone()->set_is_resizable(false),
+                                                  pool_info));
     ARM_COMPUTE_EXPECT(is_valid == expected, framework::LogLevel::ERRORS);
```
gunes-arm, line 201: Can you add functional correctness tests, too?

gunes-arm (changes requested):
fix: enable asm requantized max pooling with differing qinfo
-->
fix: Enable asm requantized max pooling with differing qinfo

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: We have asm kernels for this use-case when `qinfos` are the same. So this is correct, I'll add new tests for this in the next patchset.

See `a64_s8_nhwc_max_generic_depthfirst`

I'll tighten this to reject `q8s` avg pool when `padding=true`

src/cpu/kernels/internal/CpuPool2dAssemblyWrapperKernel.cpp
morgolock: Addressed in next patchset

We support both types. Only missing case we will reject is `q8s` when `qinfos` are different.

tests/validation/NEON/PoolingLayer.cpp
morgolock: Sure, will be reworked in next patchset.

tests/validation/NEON/PoolingLayer.cpp
morgolock: Added in next pathset

tests/validation/NEON/PoolingLayer.cpp
```diff
@@ -163,12 +185,44 @@ DATA_TEST_CASE(Validate, framework::DatasetMode::ALL, zip(
...
+TEST_CASE(ValidatePaddedAvgDifferentQuantizationInfo, framework::DatasetMode::ALL)
+{
+    TensorInfo input_info(TensorShape(3U, 15U, 11U, 1U), 1, DataType::QASYMM8, DataLayout::NHWC);
+    TensorInfo output_info(TensorShape(3U, 8U, 6U, 1U), 1, DataType::QASYMM8, DataLayout::NHWC);
+    const auto pool_info = PoolingLayerInfo(PoolingType::AVG, 3, DataLayout::NHWC, PadStrideInfo(2, 2, 1, 1), false);
+
+    input_info.set_quantization_info(QuantizationInfo(0.25f, 11));
+    output_info.set_quantization_info(QuantizationInfo(0.5f, 7));
+
+    const bool is_valid = bool(NEPoolingLayer::validate(&input_info.clone()->set_is_resizable(false),
```
gunes-arm: I don't think you need to clone here.  It was for constant objects as far as I recall.

tests/validation/NEON/PoolingLayer.cpp
```diff
@@ -448,6 +502,22 @@ FIXTURE_DATA_TEST_CASE(RunMixedDataLayout,
 }
 TEST_SUITE_END() // QASYMM8
 TEST_SUITE(QASYMM8_SIGNED)
+FIXTURE_DATA_TEST_CASE(PaddedMaxSameQInfo,
```
gunes-arm, line 525: Can you add QASYMM8 tests, too? They use different underlying kernels. Probably, we need to put arbitrary quantization infos to src and dst as they don't have to be 0 offset.

gunes-arm (approved):

gunes-arm (changes requested):
Accidental approval last time

tests/validation/NEON/PoolingLayer.cpp
morgolock: Already there in
```
455 FIXTURE_DATA_TEST_CASE(QASYMM8PaddedMax,
456                        NEPoolingLayerQuantizedFixture<uint8_t>,
457                        framework::DatasetMode::PRECOMMIT,
458                        PoolingDatasetQASYMM8PaddedMax)
459 {                                                   
460     validate(Accessor(_target), _reference, tolerance_qasymm8);
461 }               
```

tests/validation/NEON/PoolingLayer.cpp
morgolock: Removed in next patchset

tests/validation/NEON/PoolingLayer.cpp
gunes-arm: I think I commented under the wrong line. What I mean is that there are no differing quantization tests for QASYMM8. It can be minimal.

morgolock: Ok, added in next patchset

gunes-arm (approved):

Issue #1284, opened 2026-04-24
[Possible Bug] Upsample operation "Padding is not supported"


lcit:
I run into an error related to the padding for a convolution + upsample operation. I pad the convolution to obtain an output volume with the same size as the input one and upsample it to twice its size. The issue persist if I set the padding to 0.
The error is raise during the call to up0.configure.

The error:
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!

ERROR in validate_arguments src/cpu/kernels/CpuScaleKernel.cpp:357: Padding is not supported
!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Minimal code extract:
```C++
  std::unique_ptr<NEConvolutionLayer>    conv0{};
  arm_compute::NEScale                   up0{};

  const TensorShape src_shape(12, 12, 2);
  src.allocator()->init(TensorInfo(src_shape, 1, DataType::F32));

  const TensorShape weights_shape_conv0(3, 3, 2, 32);
  const TensorShape biases_shape_conv0(32);
  const TensorShape out_shape_conv0(12, 12, 32);

  weights0.allocator()->init(TensorInfo(weights_shape_conv0, 1, DataType::F32));
  biases0.allocator()->init(TensorInfo(biases_shape_conv0, 1, DataType::F32));
  out_conv0.allocator()->init(TensorInfo(out_shape_conv0, 1, DataType::F32));

  const TensorShape out_shape_up0(24, 24, 32);
  out_up0.allocator()->init(TensorInfo(out_shape_up0, 1, DataType::F32));

  conv0->configure(&src, &weights0, &biases0, &out_conv0,
                   PadStrideInfo(1 /* stride_x */, 1 /* stride_y */, 1 /* pad_x */, 1 /* pad_y */));

  up0.configure(&out_conv0, &out_up0, ScaleKernelInfo(InterpolationPolicy::NEAREST_NEIGHBOR, 
                                                     BorderMode::UNDEFINED));
```

Version and configuration: 
v53.0.0
Linux pynq 5.4.0-xilinx-v2020.1 #1 SMP PREEMPT Thu Apr 8 18:22:30 UTC 2021 armv7l armv7l armv7l GNU/Linux

Build:
```
CXX="${BASE}gcc-linaro-7.5.0-2019.12-x86_64_arm-linux-gnueabihf/bin/"
[...]

PR #1288, opened 2026-05-08
fix: default ScaleKernelInfo to no padding

morgolock:
The short ScaleKernelInfo(policy, border_mode) form implicitly enabled use_padding, but the CPU/NEON scale kernel rejects padding as an unsupported configuration. That made NEScale::configure() fail unless callers explicitly overrode the default.

Change the default use_padding value to false so the public API matches the supported behavior, and add validation coverage for the default ScaleKernelInfo path.

This solves https://github.com/ARM-software/ComputeLibrary/issues/1284

Change-Id: I86645004f33234268fb8257147a1dc7a9a62637e


Dongsung-arm (approved):
Looks good to me. It matches what the CPU/NEON kernel supports and avoids unexpected failures.

PR #1289, opened 2026-05-12
feat: Expose accumulation mode flag via Conv2dInfo

Sqvid:
Users of the functional and experimental operator convolution APIs, e.g, `arm_compute::NEGEMMConv2d`, or `arm_compute::experimental::op::CpuGemmDirectConv2d`, can make use of `fp32` accumulation by setting this flag in `Conv2dInfo` during the `validate()` and `configure()` steps.

Commit 5e40456 changed the default behaviour of `CpuGemmDirectConv2d` to accumulate in f32 unless enable_fast_math was set. However, this can produce regressions for users expecting the old behaviour. This change exposes the flag to user directly, making fp32 accumulation opt-in.

Change-Id: I3203bdbbfa5152a64438941dd138bab6feb1cec2

cc: @morgolock @gunes-arm

### DIFF
--- a/arm_compute/runtime/FunctionDescriptors.h
+++ b/arm_compute/runtime/FunctionDescriptors.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2023, 2025 Arm Limited.
+ * Copyright (c) 2019-2023, 2025-2026 Arm Limited.
  *
  * SPDX-License-Identifier: MIT
  *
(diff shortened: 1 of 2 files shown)


arm_compute/runtime/FunctionDescriptors.h
```diff
@@ -83,6 +85,7 @@ struct Conv2dInfo
     bool                enable_fast_math{false};
     unsigned int        num_groups{1};
     WeightsInfo         weights_info{};
+    bool                use_fp32_acc{false};
```
gunes-arm, line 88: Can you add an inline comment saying 
// Relevant only for Fp16

Sqvid:
Closing this to rename the source branch.

PR #1297, opened 2026-06-17
feat: enable `QASYMM8_SIGNED`→`F32` in `CpuGemmAssemblyDispatch`

alvoron:
Two gaps in `CpuGemmAssemblyDispatch` blocked `QASYMM8_SIGNED` input from validating with an `F32` output tensor, even though the underlying `arm_gemm` kernel `GemmInterleaved<int8_t, int8_t, float, DequantizeFloat>` already supports this combination on AArch64.

Gap 1 - `has_opt_impl`: The `QASYMM8_SIGNED` branch only tested for `S32` and `S8` outputs. Passing `F32` fell through to the `S8`→`S8` `Requantize32` check, which fails because `has_opt_gemm<int8_t, int8_t, int8_t, Requantize32>` and `has_opt_gemm<int8_t, int8_t, float, DequantizeFloat>` are different instantiations.

Gap 2 - `validate`: There was no output-type guard for `QASYMM8_SIGNED` input at all. The equivalent guard for `QASYMM8` explicitly allowed `QASYMM8`/`S32`/`F32`; `QASYMM8_SIGNED` had no such allowance, so `F32` output reached downstream checks with no clear error.

PR #1298, opened 2026-06-17
feat: optimized QASYMM8_SIGNED->F32 direct convolution path

alvoron:
## Motivation
Inference frameworks (e.g., OpenVINO) often run int8-quantized activations with float32 output. Previously this required a multi-step chain: quantized GEMM accumulating into int32, a separate `GEMMLowp` output-stage operator, and a dequantize/cast step. This adds a single-kernel path that takes `QASYMM8_SIGNED` input and weights and writes F32 output directly, reducing memory traffic and operator overhead.

## Dependency
**Requires https://github.com/ARM-software/ComputeLibrary/pull/1297 to be merged first.**
That branch provides the `dequant_a_offset`/`dequant_b_offset` fields on `AsmGemmInfo`, the relaxed type guard in `CpuGemmAssemblyDispatch::validate` (removing the "Only S32 output" restriction for S8 input), and the `create_arm_gemm_dequant` changes that pass offsets into `DequantizeFloat` with the correct combined scale. Without it this branch does not build correctly standalone.

## Technical approach

The existing `DequantizeFloat` output stage is extended with `a_offset` and `b_offset` fields. `GemmInterleaved` is taught to:
  1. **Pack row sums of A** into the A panel (via `transforms_quantized` with multiplier = 1) when `b_offset != 0`, for per-row offset correction.
  2. **Compute column sums of B** (weights) during `set_pretransposed_B_array` when `a_offset != 0`, stored alongside the pretransposed weight buffer.
  3. **Apply all three correction terms** in `dequantize_block_32<float>` at merge time.

K-blocking is conservatively disabled for the `DequantizeFloat + MergeStep` case (matching the existing `Requantize32` policy) since row sums must cover full K.

`CpuGemmDirectConv2d` detects the `QASYMM8_SIGNED`→`F32` path by type, reads zero-points from `QuantizationInfo`, and passes them to the dispatch via `AsmGemmInfo`. `CpuConv2d::get_convolution_method` automatically routes NHWC `QASYMM8_SIGNED`→`F32` convolutions to `GEMM_CONV2D` (no explicit flag).

Also fixes a latent bug in `dequantize_block_32<float>`: the expression `val * qp.scale` (integer × float, silently truncating for large accumulators) is corrected to `static_cast<float>(val) * qp.scale`.

## Asymmetric correction formula
out[m,n] = (raw_acc[m,n]
              − a_offset · Σ_k b[k,n]     (per-column, from col sums)
              − b_offset · Σ_k a[m,k]     (per-row, from row sums)
              + a_offset · b_offset · K    (cross-term)
             ) · (scale_a · scale_b) + bias[n]

### DIFF
--- a/src/core/NEON/kernels/arm_gemm/gemm_interleaved.hpp
+++ b/src/core/NEON/kernels/arm_gemm/gemm_interleaved.hpp
@@ -281,6 +281,14 @@ void kernel_and_merge<false, false, DequantizeFloat>::run(
         offset_bias = bias + n_0;
     }
 
+    // When b_offset != 0, row sums of A are packed at the end of the A panel
+    // (appended by the quantized PrepareA transform with multiplier=1).  Read them
+    // to pass to dequantize_block_32 for per-row offset correction.
+    const int32_t *row_sum = nullptr;
+    if (dq.b_offset != 0) {
+        row_sum = reinterpret_cast<const int32_t *>(a_ptr + strategy::out_height() * kern_k);
+    }
+
     strat.kernel(// A and B pointers are just the packed panels.
                  a_ptr, b_panel,
                  // Provide relevant part of output array and row stride.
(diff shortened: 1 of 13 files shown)


morgolock:
Closing this in favor of https://github.com/ARM-software/ComputeLibrary/pull/1302

PR #1300, opened 2026-06-25
fix: Fix OpenCL(TM) C++ header include path in cl_definitions.h

morgolock:
Commit 5c4dc9469632bef3c86da6b52506863450232b28 moved the OpenCL(TM) C++ bindings include into cl_definitions.h but used the source-tree relative path "include/CL/opencl.hpp".

This breaks consumers using installed ComputeLibrary headers, such as Arm NN with OpenCL(TM) enabled, because the installed public header is available as <CL/opencl.hpp>.

Use the public include path so cl_definitions.h works both from the source tree and from installed/package layouts.

Change-Id: If381cd88f2139cfd17b696e6173c8f192fc35089


ArmDude (approved):
Looks good

PR #1302, opened 2026-07-01
feat: Enable QASYMM8_SIGNED to floating-point assembly dequantization

morgolock:
Enable QASYMM8_SIGNED input and weights to use the F32 DequantizeFloat assembly output stage, including the direct convolution selection path.

Propagate input and weight zero-points plus the unrounded mathematical K depth so asymmetric offset correction uses the real GEMM/convolution depth rather than arm_gemm's padded internal K.

Add NEON validation coverage for the direct I8S8F32 convolution path.

Performance was checked on a A76, pinned to CPU 4 with one thread. The change is neutral on large workloads and improves the small NHWC signed int8 to F32 convolution path: QASYMM8_SIGNED/RunSmallDequantizeF32 NHWC/no-activation cases show a 1.36x geomean speedup over github/main, with the tiniest single-batch cases improving by roughly 1.45x to 3.13x.

Change-Id: Ie723d3da629d48de6de737c425bf7ad48e0f7feb


src/cpu/operators/CpuGemmDirectConv2d.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2023 Arm Limited.
+ * Copyright (c) 2021, 2023-2026 Arm Limited.
```
gunes-arm: This should be 2023, 2026

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
```diff
@@ -1171,7 +1213,7 @@ void CpuGemmAssemblyDispatch::configure(
 #ifdef __aarch64__
         case DataType::U8:
         case DataType::QASYMM8:
-            if (b->data_type() == DataType::S8 || b->data_type() == DataType::QASYMM8_SIGNED)
+            if (is_quantized_int8_rhs(b->data_type()))
```
gunes-arm: I think this function is confusing from a lot of aspects. Its name is confusing, it replaces different pieces of code and adds per channel checks as well. Why not leave the conditional logic as it is?

src/cpu/operators/internal/CpuGemmAssemblyDispatch.h
gunes-arm, line 31: cstdint

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -1320,6 +1320,143 @@ class HasOptImplFixture : public framework::Fixture
...
+    {
+        switch (tensor.data_type())
+        {
+            case DataType::QASYMM8_SIGNED:
+            {
+                std::uniform_int_distribution<int32_t> dist(-127, 127);
+                library->fill(tensor, dist, i);
+                break;
+            }
+            case DataType::F32:
```
gunes-arm: We don't need this I think. Only the output is F32

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -311,6 +311,445 @@ TEST_SUITE_END() // QASYMM8_SIGNED
 
 // clang-format on
 // *INDENT-ON*
+
+#ifdef __aarch64__
+// *INDENT-OFF*
```
gunes-arm: I don't think we need these two formatter-off lines.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -311,6 +311,445 @@ TEST_SUITE_END() // QASYMM8_SIGNED
...
+// clang-format off
+TEST_SUITE(DirectI8S8F32)
```
gunes-arm: I think this should be inside QASYMM8_SIGNED suite. And, why is the name equal to DirectI8S8F32?

Why not just TEST_SUITE(DequantizeFloat)? Same goes for the fixture name.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -311,6 +311,445 @@ TEST_SUITE_END() // QASYMM8_SIGNED
...
+        make("WeightsShape", { TensorShape(3U, 3U, 32U, 16U) }),
+        make("BiasShape",    { TensorShape(16U) }),
+        make("OutputShape",  { TensorShape(6U, 6U, 16U) }),
+        make("ConvInfo",     { PadStrideInfo(2, 2, 0, 0) }),
+        make("Dilation",     { Size2D(1U, 1U) }),
+        make("ReshapeWeights", { true }),
+        make("DataLayout",   { DataLayout::NHWC }),
+        make("ActivationInfo", { ActivationLayerInfo(),
+                                 ActivationLayerInfo(ActivationLayerInfo::ActivationFunction::RELU) }),
+        make("InputQI",      { QuantizationInfo(0.25f, 0) }),
```
gunes-arm: Why not combine all offset configurations here?

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -311,6 +311,445 @@ TEST_SUITE_END() // QASYMM8_SIGNED
...
+                                 ActivationLayerInfo(ActivationLayerInfo::ActivationFunction::RELU) }),
+        make("InputQI",        { QuantizationInfo(0.25f, -20) }),
+        make("WeightsQI",      { QuantizationInfo(0.125f, 10) })
+    )
+)
+{
+    validate(Accessor(_target), _reference, rel_tolerance_f32, tolerance_num_dequantize_f32, float(abs_tolerance_f32));
+}
+
+/** Positive input offset + stride-2 + padding: exercises the row-sum computation
```
gunes-arm: All these combinations can really be achieved with `combine`. I think this is too many lines of test code.

gunes-arm (changes requested):

src/cpu/operators/CpuGemmDirectConv2d.h
morgolock: Solved in next patch

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Solved in next patch. Reverted change and removed the helper `is_quantized_int8_rhs`

src/cpu/operators/internal/CpuGemmAssemblyDispatch.h
morgolock: Added `#include <cstdint>`

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Removed code to fill F32

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done

Passavee-Losripat:
Hi, I've been testing this PR for OpenVINO's ARM CPU plugin int8 enablement and found a possible validation gap in the new QASYMM8_SIGNED->F32 dequantization path.

`QSYMM8_PER_CHANNEL` weights are correctly rejected by `validate()` on this path. However, `QASYMM8_SIGNED` weights whose `QuantizationInfo` carries a per-channel scale vector are accepted and the kernel then dequantizes every output channel with `scale[0]` (the scalar `uniform().scale` product in `create_arm_gemm_dequant`).

I create minimal repo to test this: all-ones i8 input, all-ones 3×3×3 i8 weights, bias 0, distinct per-channel scales s[c] -> each interior output should be 27·s[c]:

```text
channel    got      expected(27*s[c])    ratio
c00      0.0270        0.0270            1.0000
c01      0.0270        0.0540            0.5000
c02      0.0270        0.1080            0.2500
...
c11      0.0270       55.2960            0.0005
```

`validate()` returns OK, but every channel is computed with `s[0]`. Since quantization frameworks (NNCF, PyTorch, TF) produce per-channel weight scales by default, this input shape is what integrations like OpenVINO naturally pass, so it fails silently on real models rather than erroring out.

I am thinking of rejecting `scale().size() > 1` on this path the same way `QSYMM8_PER_CHANNEL` is rejected, unless per-channel support is planned. I can share the full standalone test program if useful.

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
```diff
@@ -996,12 +996,22 @@ Status CpuGemmAssemblyDispatch::has_opt_impl(arm_compute::WeightFormat &expected
                                                                                             {})),
                     "We could not find an optimized kernel for U8/QASYMM8 input and U32 output");
             }
-            else if (b->data_type() == DataType::QASYMM8_SIGNED)
+            else if (b->data_type() == DataType::S8 || b->data_type() == DataType::QASYMM8_SIGNED)
```
gunes-arm: Why do we add S8 data type to here?

src/cpu/operators/CpuGemmDirectConv2d.h
```diff
@@ -55,6 +55,7 @@ class CpuGemmDirectConv2d : public ICpuOperator
      * |:--------------|:--------------|:--------------|:--------------|
      * |QASYMM8        |QASYMM8        |S32            |QASYMM8        |
      * |QASYMM8_SIGNED |QASYMM8_SIGNED |S32            |QASYMM8_SIGNED |
+     * |QASYMM8_SIGNED |QASYMM8_SIGNED |F32            |F32            |
```
gunes-arm, line 58: How about Fp16? And, Do we support mixed sign inputs? Such as UInt8 inputs and Int8 weights. That's what I felt from the changes in the glue layer. I also don't see the tests for mixed sign inputs.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +306,178 @@ FIXTURE_DATA_TEST_CASE(
...
+                         TensorInfo(TensorShape(16U, 3U, 3U, 8U), 1, DataType::QASYMM8_SIGNED, DataLayout::NHWC),
+                         TensorInfo(TensorShape(16U, 3U, 3U, 8U), 1, DataType::QASYMM8_SIGNED, DataLayout::NHWC)}),
+                   make("BiasInfo",
+                        {TensorInfo(TensorShape(8U), 1, DataType::F32), TensorInfo(TensorShape(16U), 1, DataType::F32),
+                         TensorInfo(TensorShape(8U), 1, DataType::F32), TensorInfo(TensorShape(8U), 1, DataType::F32)}),
+                   make("DstInfo",
+                        {TensorInfo(TensorShape(8U, 6U, 6U), 1, DataType::F32, DataLayout::NHWC),
+                         TensorInfo(TensorShape(16U, 4U, 4U), 1, DataType::F32, DataLayout::NHWC),
+                         TensorInfo(TensorShape(8U, 6U, 6U), 1, DataType::F32, DataLayout::NHWC),
+                         TensorInfo(TensorShape(8U, 6U, 6U), 1, DataType::F32, DataLayout::NHWC)}),
```
gunes-arm, line 358: Shouldn't we make some F16 & adjust the `expected` in the body with `cpu_supports_dtypes`?

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +306,178 @@ FIXTURE_DATA_TEST_CASE(
...
+    src_info.set_quantization_info(QuantizationInfo(0.25f, src_offset));
+    wgt_info.set_quantization_info(QuantizationInfo(0.125f, wgt_offset));
+
+    const Status s = NEConvolutionLayer::validate(&src_info, &wgt_info, &bias_info, &dst_info,
+                                                  PadStrideInfo(1, 1, 0, 0), WeightsInfo(), Size2D(1U, 1U),
+                                                  ActivationLayerInfo(), false /*fast_math*/, 1 /*num_groups*/);
+    ARM_COMPUTE_EXPECT(bool(s) == expected, framework::LogLevel::ERRORS);
+}
+
+/** Verify GEMM_CONV2D is returned automatically for NHWC QASYMM8_SIGNED to F32 dequantized convolution. */
```
gunes-arm: Float

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +306,178 @@ FIXTURE_DATA_TEST_CASE(
...
+TEST_CASE(GetConvMethod, framework::DatasetMode::ALL)
+{
+    const QuantizationInfo qi(0.25f, 0);
+    TensorInfo             src_info(TensorShape(16U, 8U, 8U), 1, DataType::QASYMM8_SIGNED, DataLayout::NHWC);
+    TensorInfo             wgt_info(TensorShape(16U, 3U, 3U, 8U), 1, DataType::QASYMM8_SIGNED, DataLayout::NHWC);
+    TensorInfo             dst_info(TensorShape(8U, 6U, 6U), 1, DataType::F32, DataLayout::NHWC);
```
gunes-arm: I'd say, loop over F32 and F16 and back with `cpu_supports_dtypes` function.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +306,178 @@ FIXTURE_DATA_TEST_CASE(
...
+using NEDequantizeFloatConvolutionFixture = DequantizeFloatConvolutionFixture<Tensor, Accessor, NEConvolutionLayer>;
+
+/** Small 3x3 convolutions with dequantized F32 output.
+ *  Shapes are in NCHW format [W, H, C] for input/output, [Kw, Kh, Cin, Cout] for weights.
+ *  InputShape/WeightsShape/BiasShape/OutputShape are zipped; each row is one consistent config. */
+FIXTURE_DATA_TEST_CASE(RunSmall,
+                       NEDequantizeFloatConvolutionFixture,
+                       framework::DatasetMode::ALL,
+                       combine(zip(make("InputShape", {TensorShape(8U, 8U, 16U), TensorShape(16U, 16U, 32U)}),
+                                   make("WeightsShape", {TensorShape(3U, 3U, 16U, 8U), TensorShape(3U, 3U, 32U, 16U)}),
```
gunes-arm: Could we add another conv size other than 3x3? As far as I can see we only test 3x3 and 1x1.

Or better, why don't we directly use something like `SmallConvolutionLayerDataset`?

gunes-arm (changes requested):

tests/validation/NEON/ConvolutionLayer.cpp
gunes-arm: By using a pre-defined dataset like `SmallConvolutionLayerDataset`, I'm talking about removing all the different combs below, like padded, strided etc. Not just for this particular suite of tests. I think it'll reduce the test code we maintain. Please let me know if we're missing any coverage by not using those pre-defined datasets.

Also, I'd like to repeat my comments about Fp16 output here.

morgolock:
> Hi, I've been testing this PR for OpenVINO's ARM CPU plugin int8 enablement and found a possible validation gap in the new QASYMM8_SIGNED->F32 dequantization path.
> 
> `QSYMM8_PER_CHANNEL` weights are correctly rejected by `validate()` on this path. However, `QASYMM8_SIGNED` weights whose `QuantizationInfo` carries a per-channel scale vector are accepted and the kernel then dequantizes every output channel with `scale[0]` (the scalar `uniform().scale` product in `create_arm_gemm_dequant`).
> 
> I create minimal repo to test this: all-ones i8 input, all-ones 3×3×3 i8 weights, bias 0, distinct per-channel scales s[c] -> each interior output should be 27·s[c]:
> 
> ```
> channel    got      expected(27*s[c])    ratio
> c00      0.0270        0.0270            1.0000
> c01      0.0270        0.0540            0.5000
> c02      0.0270        0.1080            0.2500
> ...
> c11      0.0270       55.2960            0.0005
> ```
> 
> `validate()` returns OK, but every channel is computed with `s[0]`. Since quantization frameworks (NNCF, PyTorch, TF) produce per-channel weight scales by default, this input shape is what integrations like OpenVINO naturally pass, so it fails silently on real models rather than erroring out.
> 
> I am thinking of rejecting `scale().size() > 1` on this path the same way `QSYMM8_PER_CHANNEL` is rejected, unless per-channel support is planned. I can share the full standalone test program if useful.

Hi @Passavee-Losripat 

Good catch. I added validation to reject QASYMM8_SIGNED weights with multiple scales on this dequantized path, since the current output stage only supports a single scalar scale.

Please try the latest patchset and let us know if this works for you.

Hope this helps

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Good catch, this is not needed and will be removed in next patchset

src/cpu/operators/CpuGemmDirectConv2d.h
morgolock: Good point. For CpuGemmDirectConv2d, this PR only exposes QASYMM8_SIGNED input + QASYMM8_SIGNED weights + F32 bias producing F32 output. 

I removed the mixed-sign uint8_t + int8_t -> F32 DequantizeFloat assembly selector/configure path so the glue layer no longer implies direct mixed-sign support, and added validation coverage showing the direct-conv path rejects both FP16 output and mixed QASYMM8 input with QASYMM8_SIGNED weights. The lower-level signed GEMMLowp FP16 path remains intact because it is separately covered by existing GEMMLowp validation.

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: I added an explicit F16 bias/output validation

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Good point, done.

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Changed F32 to Float in the comment

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: I replaced the hand-written RunSmall/RunStride2/Run1x1/RunPadded combinations with SmallConvolutionLayerDataset

src/cpu/operators/CpuConv2d.cpp
```diff
@@ -118,8 +118,13 @@ Status CpuConv2d::validate(const ITensorInfo         *input,
 {
     ARM_COMPUTE_TRACE_EVENT(ARM_COMPUTE_PROF_CAT_CPU, ARM_COMPUTE_PROF_LVL_CPU, "CpuConv2d::validate");
     ARM_COMPUTE_RETURN_ERROR_ON_MSG((num_groups != 1), "Grouping (num_groups != 1) is not supported on Neon");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(
+        input->data_type() == DataType::QASYMM8_SIGNED && weights->data_type() == DataType::QASYMM8_SIGNED &&
+            output->data_type() == DataType::F32 && weights->quantization_info().scale().size() > 1,
+        "Per-channel QASYMM8_SIGNED weight scales are not supported for dequantized F32 convolution");
```
gunes-arm: Quite confused here. When we have per channel quantized weights, the data type should be provided as `QSYMM8_PER_CHANNEL`. This should be the correct usage assumption throughout the library and how we validate against. We may need to think about trying to prevent this misuse in another way.

@Passavee-Losripat would you mind sharing your reproducible? And, do we have any workloads/integration points that OpenVino relies on ACL accepting a per-channel quantized weights tensor with QASYMM8_SIGNED/QASYMM8? Because, to me, we might prevent some `TensorInfo` being constructed this way, as it's not the correct usage of the types, but that's going to cause crashes instead of silently validating inside a function/operator.

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
```diff
@@ -1020,6 +1021,22 @@ Status CpuGemmAssemblyDispatch::has_opt_impl(arm_compute::WeightFormat &expected
                                                                                          {})),
                     "We could not find an optimized kernel for S8/QASYMM8_SIGNED input and S32 output");
             }
+            else if (a->data_type() == DataType::QASYMM8_SIGNED && d->data_type() == DataType::F32)
```
gunes-arm: There is a bit of a weird situation here. 

I'd remove this check because that's what the switch/case is for and remove the S8 case from the switch. Would we breaking anything? I'm not too sure. Only if there is some misuse of data types. Same situation goes with comments and U8 type.

So, I'd suggest
- Ignore S8 was in the case statement above and just remove `a->data_type() == DataType::QASYMM8_SIGNED` from the ifs here as it's already checked.
- Create a ticket to resolve the S8/U8 confusion with tests involved.

src/cpu/operators/CpuGemmDirectConv2d.h
gunes-arm: Regarding Fp16 with regular Int8 inputs, I can see it delegates the validation to arm_gemm here:

https://github.com/ARM-software/ComputeLibrary/pull/1302/changes#diff-83002887a0d8a4d2a4abe6c0ddb3c2133695a3dfbe00ca06ffb1b75c746fdf16R1031

I'm on board your with mixed-sign comment. We don't support it as it seems.

tests/validation/NEON/ConvolutionLayer.cpp
gunes-arm: If we end up supporting Fp16 based on the above discussions, we need to add a `true` case.

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +307,141 @@ FIXTURE_DATA_TEST_CASE(
...
+        TensorInfo wgt_info(TensorShape(16U, 3U, 3U, 8U), 1, DataType::QASYMM8_SIGNED, DataLayout::NHWC);
+        TensorInfo dst_info(TensorShape(8U, 6U, 6U), 1, data_type, DataLayout::NHWC);
+        src_info.set_quantization_info(qi);
+        wgt_info.set_quantization_info(qi);
+
+        const ConvolutionMethod m = NEConvolutionLayer::get_convolution_method(
+            &src_info.clone()->set_is_resizable(true), &wgt_info.clone()->set_is_resizable(true),
+            &dst_info.clone()->set_is_resizable(true), PadStrideInfo(1, 1, 0, 0), WeightsInfo(), Size2D(1U, 1U),
+            ActivationLayerInfo(), false /*fast_math*/);
+        const bool expected_direct_conv = data_type == DataType::F32 && cpu_supports_dtypes({data_type});
```
gunes-arm: Kind of unnecessary to check cpu support here as it's fp32 only based on the first check

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +307,141 @@ FIXTURE_DATA_TEST_CASE(
...
+        ARM_COMPUTE_EXPECT((m == ConvolutionMethod::GEMM_CONV2D) == expected_direct_conv, framework::LogLevel::ERRORS);
+    }
+}
+
+TEST_CASE(ValidateRejectsPerChannelWeightScales, framework::DatasetMode::ALL)
```
gunes-arm, line 406: We'll need to revisit this based on how the per channel execution should be done.

gunes-arm (changes requested):

src/cpu/operators/CpuConv2d.cpp
Passavee-Losripat: Hi, @gunes-arm and @morgolock. I attached [acl_per_channel_dequant_repro.cpp](https://github.com/user-attachments/files/30323981/acl_per_channel_dequant_repro.cpp) here as a self-contained repro: all-ones i8 input, all-ones 3\*3\*3 i8 weights, zero bias, distinct per-channel scales, so each interior output must equal 27\*s[c] with no reference implementation needed. I re-ran it against the latest patchset and the per-channel vector case now fails `validate()` (raised from `CpuConv2d::validate`), and the per-tensor scalar case still validates and computes bit-exactly, so the added check resolves the issue I reported without over-rejecting.

On your question about integration points, OpenVINO relies on ACL accepting `QASYMM8_SIGNED` weights with a per-channel scale vector today, in two places:
1. The int8 FullyConnected path that is already merged in OpenVINO master (`src/plugins/intel_cpu/src/nodes/executors/acl/acl_lowp_fullyconnected.cpp`) constructs exactly this shape: `QASYMM8_SIGNED` weights with `QuantizationInfo(dequantizationScales)` where the scale vector is per-output-channel.
2. The int8 Convolution enablement in progress (which this PR would serve) follows the same pattern.

On the dtype question, OpenVINO's quantization pipeline produces symmetric per-channel i8 weights (zero-point 0), which semantically maps to `QSYMM8_PER_CHANNEL` but we got two reasons why we currently used `QASYMM8_SIGNED + vector` instead:
- The scale vector OpenVINO attaches is not strictly the weights' own quantization scale but it's the combined dequantization scale (src_scale \* weight_scale, per output channel, folded together by our graph-level transformations), with the src tensor's `QuantizationInfo` set to trivial (scale=1) to compensate. It's effective scale attached to the weights tensor rather than a literal property of the weights data.
- On the standard requantize path, this already works correctly: `quantization::compute_quantized_multipliers_and_shifts` (`src/core/utils/quantization/AsymmHelpers.cpp`) sizes its multiplier array from `weights->quantization_info().scale().size()` and iterates every entry, with no dtype check at all. So the library's own output-stage math already reads and applies every entry in the scale vector, regardless of whether the tag is `QSYMM8_PER_CHANNEL` or `QASYMM8_SIGNED`. The FC integration above already relies on exactly this behavior, merged in OpenVINO master today.

Because of that, rejecting this combination when constructing `TensorInfo` could break an already-working OpenVINO integration on the standard requantization path, where the per-channel vector is handled correctly. I think new rejection at `validate()` looks quite right to me. It would reject only the output stage that cannot actually consume the vector, while allowing the existing supported path to continue working. It also turns the current silent incorrect result into a clear validation failure at the point where the unsupported combination is used.

Happy to help test further if useful.

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Thanks, agreed. I’ll keep this PR scoped to the QASYMM8_SIGNED -> F32 path, remove the redundant a->data_type() check inside the QASYMM8_SIGNED case, and avoid changing the S8/U8 handling here. I’ll create a follow-up MLCE/Jira ticket for the broader S8/U8 vs QASYMM8/QASYMM8_SIGNED validation cleanup with dedicated tests.

src/cpu/operators/CpuGemmDirectConv2d.h
morgolock: Thanks, I investigated the FP16 path in more detail.

arm_gemm already has a signed int8 DequantizeFloat FP16 path, so I updated the ACL glue/tests to treat this as QASYMM8_SIGNED input + QASYMM8_SIGNED weights -> floating-point output, covering both F32 and F16. Mixed-sign remains unsupported and out of scope.

While checking this, I found that dequantize_block_32<__fp16>() had the extended col_bias/row_sum/k_total arguments but was not applying the asymmetric zero-point corrections. I fixed it to mirror the F32 path:

  -a_offset * col_sum
  -b_offset * row_sum
  +a_offset * b_offset * K

I also added an F16 runtime validation case with non-zero input and weight zero-points, so the asymmetric correction path is covered.

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Addressed

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Addressed

tests/validation/NEON/ConvolutionLayer.cpp
```diff
@@ -306,11 +307,173 @@ FIXTURE_DATA_TEST_CASE(
...
+                               make("ReshapeWeights", {true}),
+                               make("DataLayout", {DataLayout::NHWC}),
+                               DequantizeFloatBasicActivationDataset,
+                               DequantizeFloatOffsetsDataset))
+{
+    validate(Accessor(_target), _reference, rel_tolerance_f32, tolerance_num_dequantize_f32, float(abs_tolerance_f32));
+}
+
+#if defined(ARM_COMPUTE_ENABLE_FP16)
+FIXTURE_DATA_TEST_CASE(RunF16NonZeroOffsets,
```
gunes-arm, line 446: I think we should use `SmallConvolutionLayerDataset` here as well.

tests/validation/fixtures/ConvolutionLayerFixture.h
```diff
@@ -1320,6 +1320,161 @@ class HasOptImplFixture : public framework::Fixture
...
+        fill(wgt_q, 1 + _hash);
+        fill(bias, 2 + _hash);
+
+        SimpleTensor<float> bias_f{bias_shape, DataType::F32};
+        for (int i = 0; i < bias.num_elements(); ++i)
+        {
+            bias_f[i] = static_cast<float>(bias[i]);
+        }
+
+        auto conv = reference::convolution_layer<int8_t, int8_t, float, float>(
```
gunes-arm: Why don't we instantiate and use a convolution_layer with halfs?

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
gunes-arm: I think we inadvertently added `S8` handling as well. Could you remove that?

src/cpu/operators/CpuConv2d.cpp
gunes-arm: Hi @Passavee-Losripat, I agree that breaking what's already working is undesirable and we can keep the validation checks introduced in this PR because they are not harmful to current functionality in any sense.

But, for the per-channel quantized tensors, we expect the data type to be `QSYMM8_PER_CHANNEL`. Although this coupling is not ideal,  for us it's too late to change this :) So, using `QASYMM8_SIGNED` or any other type with a quantization scale vector of length > 1 is an API misuse we couldn't prevent.

Many different places in the library rely on this type. We check `is_data_type_quantized_per_channel` in both Cpu and Gpu backends, including matrix multiplicaton, convolution and depthwise routines. So, I expect we can find more cases where validation doesn't fail but functional correctness does if the types aren't used properly. Also, having these validations in only some places is another validation gap for us as it seems like we are accepting length > 1 quantization info with QASYMM8_SIGNED in other operators.

As I said, I think the current set of additional checks can be added, but I recommend using `QSYMM8_PER_CHANNEL` where applicable. Could you check if this is achievable for a future release so that we can add guards in `TensorInfo` as well?

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
```diff
@@ -1130,6 +1161,19 @@ Status CpuGemmAssemblyDispatch::validate(
...
             (d->data_type() != DataType::QASYMM8 && d->data_type() != DataType::S32 && d->data_type() != DataType::F32),
         "Only QASYMM8/S32/F32 output supported for QASYMM8 input");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(a->data_type() == DataType::QASYMM8 && b->data_type() == DataType::QASYMM8_SIGNED &&
+                                        d->data_type() == DataType::F32,
+                                    "QASYMM8 input with QASYMM8_SIGNED weights and F32 output is not supported");
+    ARM_COMPUTE_RETURN_ERROR_ON_MSG(a->data_type() == DataType::QASYMM8_SIGNED &&
+                                        (d->data_type() != DataType::QASYMM8_SIGNED &&
+                                         d->data_type() != DataType::S32 && d->data_type() != DataType::F32 &&
+                                         d->data_type() != DataType::F16),
+                                    "Only QASYMM8_SIGNED/S32/F32/F16 output supported for QASYMM8_SIGNED input");
```
gunes-arm, line 1168: Can we create a ticket for adding some runtime validation checks to TensorInfo class so that it throws an exception if per-channel quantized data type is not used together with length > 1 quantization info?

And, after that, can we add a code comment just before every validation we did for temporarily handling this case, saying that these checks should be removed once the mentioned ticket is resolved? 

It'd be a good idea if we also add to the desc. of the ticket that we'd like these comments gone as part of DoD.

gunes-arm (changes requested):
Can you make the commit & PR title start with uppercase letter? Thanks :)

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Agreed. I’ll create a follow-up ticket for central TensorInfo validation of this metadata consistency issue.

For this PR, I’ll keep the local guards as temporary protection for the DequantizeFloat path and add TODO comments before them referencing the ticket. I’ll also add to the ticket DoD that these temporary comments/checks should be removed once TensorInfo rejects invalid per-channel quantization metadata centrally.

Edit: I created COMPMID-9268

morgolock:
> Can you make the commit & PR title start with uppercase letter? Thanks :)

Updated the commit title to start with an uppercase letter. I will update the PR title to match.

tests/validation/NEON/ConvolutionLayer.cpp
morgolock: Done. I changed the F16 runtime coverage to reuse SmallConvolutionLayerDataset as well, while keeping a non-zero-offset combination so the asymmetric correction path is still exercised.

tests/validation/fixtures/ConvolutionLayerFixture.h
morgolock: Good point. The F16 fixture now instantiates the reference convolution with half bias/output types instead of always using a float reference path.

src/cpu/operators/CpuGemmConv2d.h
gunes-arm, line 81: This table should be updated, too.

src/runtime/NEON/functions/NEConvolutionLayer.cpp
gunes-arm, line 1: Can you have a look which public functions' data type tables we should update? Clearly NEConvolutionLayer is one of them. There may be other public function that directly use CpuConv2d, CpuGemmConv2d etc.

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
gunes-arm: Can you remove the added logic for `case DataType::S8` in both validate & configure? I think it sends the wrong message as if we support `S8`, but we shouldn't. Since they were there before, I'd leave the case statement as is, but I wouldn't add any logic below that case statement and create a ticket to remove S8 and U8 from this switch case. Especially, if we're not breaking anything, we could just do that in a single PR. If we're breaking anything, we can deprecate and say this is not the right type to use.

src/runtime/NEON/functions/NEConvolutionLayer.cpp
morgolock: Done. I searched the convolution public/operator dtype tables and updated the relevant CpuConv2d/CpuGemmConv2d/NEConvolutionLayer documentation to include the new QASYMM8_SIGNED input/weights -> F32/F16 output combinations. I kept the tables scoped to signed quantized input/weights only; mixed-sign and plain S8 remain undocumented and unsupported.

src/cpu/operators/CpuGemmConv2d.h
morgolock: Addressed

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Done. I kept the existing S8 case labels untouched, but removed the new DequantizeFloat/F32/F16 logic that had been added under the S8 path. This PR now stays scoped to QASYMM8_SIGNED input/weights with F32/F16 output. The broader S8/U8 vs QASYMM8/QASYMM8_SIGNED cleanup should be handled separately with dedicated tests.

gunes-arm: But, you've added `a->data_type() == DataType::QASYMM8_SIGNED ` checks back 🤔 See my initial message in the thread.

src/cpu/operators/CpuConv2d.cpp
gunes-arm: Hi @Passavee-Losripat, any thoughts on the timeline for fixing the data type usage for per-channel quantized models in OpenVino?

Passavee-Losripat: Hi @gunes-arm, sorry for the slow reply. I wanted to check with the team first.

We’re planning to move the per-channel weight tensors to `QSYMM8_PER_CHANNEL` in both places where we currently use `QASYMM8_SIGNED` with a multi-value scale vector. I’ll be leading this work and plan to start in the first week of September. I also tested this PR's latest patchset against our downstream OpenVINO usage (our ARM CPU plugin's `Conv+FQ` test suite with int8-quantized model inference) and confirmed it doesn't cause any new problems for us today.

Thanks for keeping the validation guards in this PR as a temporary safeguard. We’re also fine with the TensorInfo-level ticket whenever you’re ready to open it. Once the changes are merged, I’d be happy to help test it against both paths.

Passavee-Losripat (approved):
I tested this PR against our downstream OpenVINO usage (ARM CPU plugin's Conv+FQ subgraph test suite and an int8-quantized YOLO model). There're no errors so far.

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
morgolock: Removed `a->data_type() == DataType::QASYMM8_SIGNED`

gunes-arm: Thank you, do we have a ticket for 
> Create a ticket to resolve the S8/U8 confusion with tests involved.

And, can we add it as a comment to the code next to ::S8 mentions?

morgolock: Thanks, I created COMPMID-9279

src/cpu/operators/internal/CpuGemmAssemblyDispatch.cpp
```diff
@@ -998,10 +1001,22 @@ Status CpuGemmAssemblyDispatch::has_opt_impl(arm_compute::WeightFormat &expected
...
+                if (d->data_type() == DataType::F32)
+                {
+                    ARM_COMPUTE_RETURN_ERROR_ON_MSG(
+                        !(arm_gemm::has_opt_gemm<uint8_t, int8_t, float, arm_gemm::DequantizeFloat>(
+                            arm_gemm_expected_wf, args, {})),
+                        "We could not find an optimized kernel for QASYMM8 input with QASYMM8_SIGNED weights and F32 "
+                        "output");
+                }
+                else
+                {
```
gunes-arm, line 1013: I'm not sure if this is guarded anywhere else, but I also suggest to add smth like
`ARM_COMPUTE_RETURN_ERROR_ON_MSG(d->data_type() != DataType::QASYMM8)` so we do not query arm_gemm `TO=uint8_t` with another type.

gunes-arm: Never mind this, I think this is checked elsewhere

gunes-arm (approved):

PR #1305, opened 2026-07-29
fix: Enable MSVC support for AArch64 code paths and Windows build fixes

Harishmcw:
### Summary

This PR improves MSVC compatibility for Windows on Arm64 by extending AArch64-specific code paths to also recognize `_M_ARM64`, and fixes the SCons build system to work correctly on native ARM64 Windows hosts.

### Motivation

- While building OpenVINO for Windows on Arm64 using MSVC, several compilation failures and runtime crashes were encountered in ARM Compute Library due to ARM64-specific code being guarded only by __aarch64__.
- MSVC defines `_M_ARM64` instead of __aarch64__, causing those code paths to be excluded.
This resulted in compilation errors such as:

```
error C2039: 'NEReorderLayer': is not a member of 'arm_compute'
error C3646: unknown override specifier
error C4430: missing type specifier
```
- Additionally, the SCons build fails on native ARM64 Windows hosts because the assembler command template defaults to POSIX-style flags (-o) instead of MSVC-style (/Fo) that clang-cl expects.

### Fix
This PR updates ARM64-specific code paths to support both GCC/Clang and MSVC by replacing architecture guards of the form:

`#ifdef __aarch64__`

with

`#if defined(__aarch64__) || defined(_M_ARM64)`

The changes include:

- ActivationLayerInfo.h — Enables ARM64-specific members and introduces an MSVC-compatible float16_t alias.
- LUTManager.h — Enables ARM64 lookup table definitions for MSVC.
- QuantizationInfo.h — Applies the ARM64 rounding policy for MSVC builds.
- NEReorderLayer.h — Makes NEReorderLayer available for _M_ARM64.

### **SConstruct**

- Explicitly sets `ASCOM `and `ASPPCOM `for the Windows build. The `os == 'windows'` block already overrides AS to clang-cl, but doesn't set the command template.
- On native ARM64 Windows hosts, SCons fails to auto-detect MSVC and falls back to POSIX-style flags (-o), which clang-cl rejects. This fix completes the existing Windows overrides.


Harishmcw:
Hi, just a quick ping on this PR. Could you please take a look when you get a chance?

Thanks!

morgolock:
Hi @Harishmcw 

I'm confused, are you trying to build with **MSVC/ cl.exe** ? If so, this is not supported, building ACL natively on Windows requires **clang**

Harishmcw:
Hi @morgolock,

Yes, we are building ACL itself using `clang-cl` via SCons, not trying to build ACL with MSVC.

The issue is that ACL is used as a dependency by OpenVINO, which is built with MSVC (`cl.exe`). When OpenVINO includes ACL's public headers, they are parsed by the MSVC compiler. MSVC defines `_M_ARM64` rather than `__aarch64__`, so the AArch64-specific declarations in the headers are skipped by the preprocessor. This causes the downstream OpenVINO MSVC build to fail because it cannot see the declarations exposed by the `clang-cl`-compiled ACL library.

The changes in this PR are to make the ACL headers compatible with this use case.

morgolock:
> Hi @morgolock,
> 
> Yes, we are building ACL itself using `clang-cl` via SCons, not trying to build ACL with MSVC.
> 
> The issue is that ACL is used as a dependency by OpenVINO, which is built with MSVC (`cl.exe`). When OpenVINO includes ACL's public headers, they are parsed by the MSVC compiler. MSVC defines `_M_ARM64` rather than `__aarch64__`, so the AArch64-specific declarations in the headers are skipped by the preprocessor. This causes the downstream OpenVINO MSVC build to fail because it cannot see the declarations exposed by the `clang-cl`-compiled ACL library.
> 
> The changes in this PR are to make the ACL headers compatible with this use case.

Thanks, that makes sense.

If ACL itself is still built with `clang-cl`, and OpenVINO is built with `cl.exe` and just consumes the ACL headers/library, then adding `_M_ARM64` to the public header guards looks reasonable.

I only have two things I'd still like to check:

1. For MSVC we're doing:

```cpp
using float16_t = uint16_t;
```

Is `LookupTable65536` (or anything containing it) ever passed across the OpenVINO/ACL boundary? If it's only there so MSVC can parse the header, that's probably fine, but if it crosses the ABI boundary I'd like to make sure using `uint16_t` here is actually safe.

2. For `ASPPCOM`, we're replacing the SCons command with:

```python
env['ASPPCOM'] = '$AS $ASPPFLAGS /c /Fo$TARGET $SOURCES'
```

Don't we also need to keep the normal preprocessing flags like `$CPPFLAGS`, `$_CPPDEFFLAGS` and `$_CPPINCFLAGS`? Otherwise we might break `.S` files that rely on defines or include paths.

Other than those two points, the approach makes sense to me.

Harishmcw:
Thanks for the review.

1. For `LookupTable65536`, OpenVINO uses `ActivationLayerInfo`, but it does not use or populate the `_lut_fp16 / LookupTable65536` member. The `uint16_t` fallback is needed for MSVC to parse the ACL public header on Windows ARM64; the LUT itself is not used by OpenVINO across the ACL boundary.
2. For `ASPPCOM`, I’ve updated the command to preserve the normal preprocessing flags (`$CPPFLAGS`, `$_CPPDEFFLAGS`, and `$_CPPINCFLAGS`) while retaining `/Fo$TARGET` for the clang-cl output.

morgolock:
Hi @Harishmcw 
Thanks for your contribution, I've triggered the precommit checks.

morgolock (changes requested):
Could you please :

- change the commit message and use the commitizen format.
      fix: Enable MSVC support for AArch64 code paths and Windows build fixes.
   Also update the PR title to match the commit message.
- Squash the two commits

Harishmcw:
Hi @morgolock,

I’ve updated the commit message and PR title to the requested format and squashed the two commits into one.

morgolock:
Thanks for addressing the comments, triggered now.

morgolock:
Hi @Harishmcw 

Thanks for the contribution. The precommit is currently failing the `Check Signed-off-by` check.

ACL requires each commit to include a DCO sign-off. Could you please amend the commit with:

```bash
git commit --amend --no-edit -s
git push --force-with-lease
```

That should add a trailer like:

```text
Signed-off-by: Your Name <your.email@example.com>
```

Once pushed, the precommit check should pass.

Harishmcw:
I’ve just done the same. Could you please retrigger the check now?

morgolock:
Thanks. Triggered now.

morgolock (approved):

PR #1308, opened 2026-09-08
fix: Avoid SVE codegen in SME wrapper

morgolock:
The arm_gemm SME interleave wrappers are instantiated from the SVE source bucket so explicit SME and Streaming-SVE assembly can be assembled. That also allowed the compiler to auto-vectorize ordinary C++ wrapper code into non-streaming SVE instructions, which is invalid on CPUs that expose SME/SME2 without normal SVE.

Split the affected wrapper into a protected SVE bucket. It keeps the existing SVE assembler target but disables compiler vectorization for the ordinary wrapper code. Normal SVE and SVE2 buckets stay unchanged, so multi-ISA builds still carry their runtime-dispatched implementations.

Resolves MLCE-2015

Change-Id: Idea17eef5a0cd422c2510ba4dbe6882f33120591


CMakeLists.txt
gunes-arm: 2026

BUILD.bazel
gunes-arm: 2026

filedefs.json
gunes-arm, line 1: You need to update the year for `filedefs.json` in `REUSE.toml`

gunes-arm (changes requested):

gunes-arm (approved):

PR #1314, opened 2026-09-14
fix: Restore exhaustive S8/U8 destination handling in `CpuGemmAssemblyDispatch`

alvoron:
a281025b89 replaced the final unconditional else of the `case DataType::S8: case DataType::QASYMM8_SIGNED:` arm in
`CpuGemmAssemblyDispatch::has_opt_impl()` with a narrower `else if (d->data_type() == DataType::QASYMM8_SIGNED)`:
```diff
-            else
+            else if (d->data_type() == DataType::QASYMM8_SIGNED)
             {
                 ARM_COMPUTE_RETURN_ERROR_ON_MSG(
                     !(arm_gemm::has_opt_gemm<int8_t, int8_t, int8_t, arm_gemm::Requantize32>(...)),
-                    "We could not find an optimized kernel for S8 input and S8 output");
+                    "We could not find an optimized kernel for QASYMM8_SIGNED input and QASYMM8_SIGNED output");
             }
```
That arm is entered for both `S8` and `QASYMM8_SIGNED` inputs, but the new chain only covers `S32`, `F32`, `F16` and `QASYMM8_SIGNED` destinations. With a plain `S8` destination no branch executes at all and control falls straight out of the `switch` into:
```cpp
    expected_weight_format = assembly_utils::map_to_arm_compute_weight_format(arm_gemm_expected_wf);
    return Status{};   // success — without ever querying arm_gemm
```

So `has_opt_impl()` reports success while never calling `has_opt_gemm()` and never assigning `arm_gemm_expected_wf`, which is returned to the caller unchanged (i.e. whatever the caller seeded it with - typically `UNSPECIFIED` or `ANY`).

That is dangerous, because `has_opt_impl()` is a public static that callers invoke directly, without going through `validate()`, to decide how to repack weights.
`UNSPECIFIED`/`ANY` carry no interleave/block counts, so `arm_compute::interleave_by()` and `block_by()` both return `0`. A caller that feeds the reported format into a weight-reordering helper (oneDNN's `acl_utils::reorder_to_weight_format()` does exactly this) computes `stride = interleave_by * block_by = 0` for every dimension, producing a zero-strided - and therefore zero-sized - weights descriptor. In OpenVINO's CPU plugin that yields a 0-byte weights buffer and a SIGSEGV at null inside the subsequent reorder, at model-compile time, on BERT INT8 models that were fine on v53.1.0.


gunes-arm:
Hi @alvoron , sign-off in the commit message is missing. Also, could you put the PR description to commit message description? The description in the commit message gets merged eventually.

alvoron:
@gunes-arm I updated the commit message and added sign-off line. Thanks!

gunes-arm (approved):

